PR #21, opened 2023-04-27
docs: add alternative install command, if the first one failed.

eu-erwin:
## Status

**READY/IN DEVELOPMENT/HOLD**

## Description

Nothing changes, just provide alternative install command on ubuntu, in case the first install command failed.

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ x ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -20,6 +20,9 @@ To install the Shorebird command line interface (CLI):
 
 ```bash
 curl --proto '=https' --tlsv1.2 https://raw.githubusercontent.com/shorebirdtech/install/main/install.sh -sSf | sh
+
+# or alternative to Linux, if the command above not working. Use bash instead of sh.
+curl --proto '=https' --tlsv1.2 https://raw.githubusercontent.com/shorebirdtech/install/main/install.sh -sSf | bash
 ```
 
 ## Windows
--- a/docs/index.md
+++ b/docs/index.md
@@ -19,9 +19,6 @@ To install the Shorebird command line interface (CLI):
 ## Mac/Linux
 
 ```bash
-curl --proto '=https' --tlsv1.2 https://raw.githubusercontent.com/shorebirdtech/install/main/install.sh -sSf | sh
-
-# or alternative to Linux, if the command above not working. Use bash instead of sh.
 curl --proto '=https' --tlsv1.2 https://raw.githubusercontent.com/shorebirdtech/install/main/install.sh -sSf | bash
 ```
 


docs/index.md
```diff
@@ -20,6 +20,9 @@ To install the Shorebird command line interface (CLI):
 
 ```bash
 curl --proto '=https' --tlsv1.2 https://raw.githubusercontent.com/shorebirdtech/install/main/install.sh -sSf | sh
+
+# or alternative to Linux, if the command above not working. Use bash instead of sh.
+curl --proto '=https' --tlsv1.2 https://raw.githubusercontent.com/shorebirdtech/install/main/install.sh -sSf | bash
```
eseidel, line 25: Thanks for the patch!

I think we just change the above line to use | bash instead of | sh.

eu-erwin: great, i assume, it is also work on mac. isn't it? If so then I can remove this PR. love your works btw.

eseidel: Can you share the logs of the failure of the first command on your machine so we can better understand what needs fixing here?

is it that `sh` doesn't exist in your path?   Or is it that `sh` has the wrong behavior?  Piping to bash seems reasonable, so a change which changed the first line to use `bash` instead of `sh` seems like a good one and we would welcome such a patch.

eu-erwin: I think the issue related to this.
[stackoverflow](https://stackoverflow.com/a/1878999)

grafik

Some output from the command.
```
user@workstation:~$ curl --proto '=https' --tlsv1.2 https://raw.githubusercontent.com/shorebirdtech/install/main/install.sh -sSf | sh
sh: 13: Syntax error: "(" unexpected (expecting "}")
```

Line 13:
```
  rc_files=("$HOME/.bashrc" "$HOME/.zshrc")
```

I am just tried replace the array like below, removing the bracket indeed works, both for `bash` and `sh`.
Can't say what the issue tho.
```
  rc_files="$HOME/.bashrc" "$HOME/.zshrc"
```

bryanoltman, line 25:
```suggestion
```bash
curl --proto '=https' --tlsv1.2 https://raw.githubusercontent.com/shorebirdtech/install/main/install.sh -sSf | bash
```

bryanoltman (approved):

PR #38, opened 2023-05-19
docs: add guide for multiple deployment tracks

felangel:
## Status

**READY**

## Description

- docs: add guide for multiple deployment tracks

### DIFF
--- a/docs/guides/multiple_deployment_tracks.md
+++ b/docs/guides/multiple_deployment_tracks.md
@@ -1,0 +1,245 @@
+---
+sidebar_position: 3
+title: 🛤️ Multiple Deployment Tracks
+description: Push patches to multiple deployment tracks
+---
+
+# Multiple Deployment Tracks
+
+This guide will walk through how to setup an app in which there are 2 deployment tracks: `internal` and `stable`. It will cover how to validate a patch on the internal track and then promote the patch to the stable track.
+
+## Prerequisites
+
+This guide assumes the Shorebird CLI is installed on your machine and that you are logged into a paid account. Refer to the [getting started](/) instructions for more information.
+
+## Create a Project
+
+:::info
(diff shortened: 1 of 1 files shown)


docs/guides/multiple_deployment_tracks.md
```diff
@@ -0,0 +1,245 @@
...
+│ shorebird_example (stable)   │ 1db93edf-4a8a-493a-9450-bd62d06cefb1 │ 1.0.0+1     │ --    │
+└──────────────────────────────┴──────────────────────────────────────┴─────────────┴───────┘
+```
+
+:::
+
+We can validate the patch by re-launching the internal release.
+
+:::warning
+If you are testing locally, DO NOT re-run `shorebird run`, instead just re-launch the app from the device or emulator directly.
```
eseidel, line 230: Why not? :). "shorebird run always builds and runs the latest code for your app.  Which means it will run with the patched code already included, rather than showing you the released code with the patch applied."?

Not sure. Or maybe we just need to make `shorebird run` give a warning in these cases?

eseidel (approved):
I think this is great.  I think guides like this also make obvious our rough edges, which gives us more incentive to polish them.

docs/guides/multiple_deployment_tracks.md
felangel: Yeah I put this in so that folks can see the patches being applied. I can reword to make this "softer" like, "If you are testing locally you do not need to re-run `shorebird run`...

eseidel: It's also fine as is.  I just like to either explain why we're asking the user to do something extra (so they can understand) or take away the footgun. :)

felangel: Yup re-worded to hopefully make it a bit more clear.

Issue #87, opened 2023-07-25
Document the flow for creating both iOS and Android from same machine.


bryanoltman:
We'll probably want to update this on:

1. [x] https://docs.shorebird.dev/code_push/release
2. [x] https://docs.shorebird.dev/code_push/patch
3. [x] https://docs.shorebird.dev/guides/code_push_quickstart
4. [x] https://docs.shorebird.dev/code_push/preview depending on where that work lands
5. [x] https://docs.shorebird.dev/guides/code_push_release
6. [ ] https://docs.shorebird.dev/architecture if @eseidel wants to talk about our Dart sdk modifications
7. [x] Update `shorebird release ios-preview` warning text.

bryanoltman:
I believe this is simply:

1. `shorebird init`
2. `shorebird release android`
3. `shorebird release ios-preview`
4. (Dart changes)
5. `shorebird patch android`
6. `shorebird patch ios-preview`

bryanoltman:
We'll probably want to update this on:

1. https://docs.shorebird.dev/code_push/release
2. https://docs.shorebird.dev/code_push/patch
3. https://docs.shorebird.dev/guides/code_push_quickstart
4. https://docs.shorebird.dev/code_push/preview depending on where that work lands
5. https://docs.shorebird.dev/guides/code_push_release (probably less urgent than the others)
6. https://docs.shorebird.dev/status to move iOS from a "coming soon" to a "usable but not production-ready" preview state.
7. https://docs.shorebird.dev/faq#what-platforms-does-shorebird-support-does-it-support-ios same as above
8. https://docs.shorebird.dev/architecture if @eseidel wants to talk about our Dart sdk modifications

eseidel:
https://github.com/shorebirdtech/docs/pull/52

PR #88, opened 2023-07-25
update init docs with fresh cli output and note about new init behavior

bryanoltman:
## Status

**READY**

## Description

Updates init to mention that `shorebird init` will add the internet permission to the android manifest if it doesn't already exist, and updates the init output to reflect changes since the doc was last updated.

Part of https://github.com/shorebirdtech/docs/issues/87


felangel (approved):
💯

PR #94, opened 2023-07-26
Update CodeMagic install instructions to support linux

bulwinkel:
## Status

**READY**

## Description

Using a linux_x2 build machine, the 'Shorebuild build' phase of the CodeMagic yaml file was failing to find the `shorebird` command on the path. I checked the output of the install script and noticed that the suggested path was different from the documentation.

Switched to mac_mini_m1 and it worked fine.

Updated the script to check OS and set path accordingly. Added the # Test at the bottom to catch errors in the path early rather than waiting until the build phase for it to fail.

This script modification has been tested on:
- mac_mini_m1
- linux_x2

### DIFF
--- a/docs/ci/codemagic.mdx
+++ b/docs/ci/codemagic.mdx
@@ -110,9 +110,22 @@ scripts:
     script: |
       # Install Shorebird
       curl --proto '=https' --tlsv1.2 https://raw.githubusercontent.com/shorebirdtech/install/main/install.sh -sSf | bash
+      
+      OS=$(uname)
 
-      # Add Shorebird to PATH
-      echo PATH="/Users/builder/.shorebird/bin:$PATH" >> $CM_ENV
+      if [ "$OS" == "Darwin" ]; then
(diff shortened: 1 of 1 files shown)


docs/ci/codemagic.mdx
```diff
@@ -110,9 +110,22 @@ scripts:
     script: |
       # Install Shorebird
       curl --proto '=https' --tlsv1.2 https://raw.githubusercontent.com/shorebirdtech/install/main/install.sh -sSf | bash
-
-      # Add Shorebird to PATH
-      echo PATH="/Users/builder/.shorebird/bin:$PATH" >> $CM_ENV
```
eseidel, line 115: The better fix might just be to use $HOME.

eseidel:
Thanks for the patch!

docs/ci/codemagic.mdx
bulwinkel: True 😀

eseidel:
Thanks for sending this!  I've made a similar change in https://github.com/shorebirdtech/docs/pull/97 per our discussion.

PR #97, opened 2023-07-26
Fix: Paths in codemagic instructions are wrong on linux.

eseidel:
As requested in https://github.com/shorebirdtech/docs/pull/94


nilsreichardt (approved):
Thanks for the fix! LGTM

When I wrote the docs, I only tested it on macOS machines.

nilsreichardt:
image

I can also confirm that it works on Linux with Codemagic 👍

eseidel:
Thanks!

bryanoltman (approved):

felangel (approved):

PR #112, opened 2023-08-09
Add instructions about escaping -- on Windows

bryanoltman:
## Status

**READY**

## Description

Due to a [bug](https://github.com/dart-lang/args/issues/252) in the args package, `--` needs to be escaped on Windows to work properly.


felangel (approved):
😢

PR #120, opened 2023-08-15
Add notes about CONTINUOUS_INTEGRATION env variable

bryanoltman:
## Status

**READY**

## Description

Removes the use of `--force`

As per [CodeMagic](https://docs.codemagic.io/yaml-basic-configuration/environment-variables/) and [GitHub](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) docs, CodeMagic will always set the `CI` and `CONTINUOUS_INTEGRATION` variables to true and GitHub will always set `GITHUB_ACTIONS` to true, so no additional setup is required once https://github.com/shorebirdtech/shorebird/pull/1099 lands.

Part of https://github.com/shorebirdtech/shorebird/issues/1097

### DIFF
--- a/docs/ci/codemagic.mdx
+++ b/docs/ci/codemagic.mdx
@@ -89,9 +89,7 @@ In the `codemagic.yaml` file, specify the `SHOREBIRD_TOKEN` under `environment`.
   <TabItem value="maclinux" label="Mac/Linux" default>
 
 1. Add the group "shorebird" to your environment groups. If you're using a
-   different group name, use that instead. Also add the `CONTINUOUS_INTEGRATION`
-   variable to skip confirmation prompts.
-
+   different group name, use that instead.
 ```yaml
 workflows:
   example:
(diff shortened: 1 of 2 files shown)


nilsreichardt (approved):
LGTM

docs/ci/codemagic.mdx
```diff
@@ -99,6 +100,9 @@ workflows:
       groups:
         # Exports the SHOREBIRD_TOKEN environment variable
         - shorebird
+      vars:
+        # Skip confirmation prompts
+        CONTINUOUS_INTEGRATION: 'true'
```
felangel: Minor but might want to rename to CI for consistency with other setups (e.g flutter, GitHub, GitLab, etc). Also if we did that users wouldn’t need to specify this value since it’s automatically set.

bryanoltman: Too slow :) Take a look at the updated comment and code

felangel (approved):

PR #128, opened 2023-08-31
docs: Write a guide on update strategies

eseidel:
This comes from frequent questions in the Discord. I'm sure
there is more we can do here.

Fixes https://github.com/shorebirdtech/shorebird/issues/1130
Fixes https://github.com/shorebirdtech/shorebird/issues/1123


bryanoltman:
Opened https://github.com/shorebirdtech/docs/pull/129 with suggestions

eseidel:
Honestly we should just land this (or something like it) and you and others should rework it to their heart's content.  I'm sure we'll get feedback from customers too.

bryanoltman (approved):

PR #129, opened 2023-09-01
update strategies edits

bryanoltman:
Had a bunch of small copy edit comments on the target branch, so thought it would be easier to just apply them myself.


eseidel:
Eek, can't see the diff with the rename.  Let me rename on the branch if it's supposed to be kebab case.

PR #133, opened 2023-09-06
docs: Add explicit mention of using Shorebird in a sideloaded env.

eseidel:
Fixes https://github.com/shorebirdtech/shorebird/issues/1178

I suspect we could write more (I initially started this as a separate
doc, but didn't have enough to say to bother splitting).


eseidel:
This may be better as a separate doc, we'll see.  The point of the separate doc is mostly about making it easy for users to find.  Increasingly users are looking for walkthroughs for their specific user journeys rather than just reference docs.  Our walkthroughs may end up repeating each other, but that's OK.

felangel (approved):

PR #148, opened 2023-09-26
Add troubleshooting section about native change warnings

bryanoltman:
## Status

**READY**

## Description

Part of https://github.com/shorebirdtech/shorebird/issues/1286. https://github.com/shorebirdtech/shorebird/issues/1286 will be closed once we add a link to this in the CLI.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -124,11 +124,12 @@ or removing the value from your export options .plist file.
 
 ## I see a `The release artifact contains native changes` warning when running `shorebird patch`, even though I haven't changed Swift/Objective-C/Kotlin/Java code
 
-The `shorebird patch` will print this warning if it detects changes to files in
-your app that correspond to native code changes (`.dex` files on Android, files
-in the `Runner.app` directory on iOS). This does not always mean that your patch
-will not work, but because shorebird cannot be sure that the changes are safe,
-and because Shorebird can't patch non-Dart code, it prints a warning.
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -122,6 +122,29 @@ version and build numbers in order to target patches at specific releases. You
...
+   use caution when publishing patches that include changes to native code from
+   plugins. In the worst case, these changes my cause your patched app to
+   crash.**
+2. A dependency/plugin produces a different output on every build. This can
+   happen if the dependency it includes a timestamp indicating when it was
+   built, for example. This kind of change is usually safe to publish, but you
+   should be sure this is the only reason you are seeing this warning.
+
+If you are confident that the changes are safe, you can ignore this warning by
+passing the `--force` flag to `shorebird patch`.
```
bryanoltman, line 146: Maybe we should add an `--ignore-native-changes` or `--ignore-change-warning` flag instead?

felangel: In addition to force?

felangel (approved):

bryanoltman:
@felangel Yeah. For the purposes of artifact diffing, it would behave as though the `--force` flag were provided, but would not skip any other checks. I'm mostly suggesting it because I don't think we want to encourage users to use `--force` in more or less any case.

vivaan-vama:
This have been publish to production ?

PR #176, opened 2024-01-19
docs(faq): using shorebird for all dart changes

felangel:
## Status

**READY**

## Description

- docs(faq): using shorebird for all dart changes

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -119,7 +119,7 @@ expect feature changes to go through store review.
 Notably Apple's App Store requires that an update "does not change the
 primary purpose of the Application by providing features or functionality
 that are inconsistent with the intended and advertised purpose of the Application."
-Also note that more Dart code changes will result in larger patch sizes.
+Also note that patch sizes correlate with the total amount of Dart changed from the original released app (each patch is a diff against the released dart code, not a diff to the previous patch).
 
 ### Does Shorebird submit to the stores for me?
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -118,7 +118,7 @@ Note that depending on how you distribute your apps, some store agreements
 expect feature changes to go through store review.
 Notably Apple's App Store requires that an update "does not change the
 primary purpose of the Application by providing features or functionality
-that are inconsistent with the intended and advertised purpose of the Application.
+that are inconsistent with the intended and advertised purpose of the Application."
 Also note that more Dart code changes will result in larger patch sizes.
 
 ### Does Shorebird submit to the stores for me?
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -111,6 +111,16 @@ Also Shorebird does not support changing native code (e.g. Java/Kotlin on
...
 attempted patch if you have changed native code.
 
+### Can I use Shorebird for all my Dart changes?
+
+Technically speaking, yes, a Shorebird patch can update any Dart code.
+Note that depending on how you distribute your apps, some store agreements
+expect feature changes to go through store review.
+Notably Apple's App Store requires that an update "does not change the
+primary purpose of the Application by providing features or functionality
+that are inconsistent with the intended and advertised purpose of the Application.
```
bryanoltman, line 121:
```suggestion
that are inconsistent with the intended and advertised purpose of the Application."
```

felangel: Good catch, thanks!

docs/faq.md
```diff
@@ -111,6 +111,16 @@ Also Shorebird does not support changing native code (e.g. Java/Kotlin on
...
+Also note that more Dart code changes will result in larger patch sizes.
```
eseidel, line 122:
```suggestion
Also note that patch sizes correlate with the total amount of Dart changed from the original released app (each patch is a diff against the released dart code, not a diff to the previous patch).
```

docs/faq.md
```diff
@@ -111,6 +111,16 @@ Also Shorebird does not support changing native code (e.g. Java/Kotlin on
 Android or Objective-C/Swift on iOS). The tool will warn you during an
 attempted patch if you have changed native code.
 
+### Can I use Shorebird for all my Dart changes?
+
+Technically speaking, yes, a Shorebird patch can update any Dart code.
```
bryanoltman, line 116:
```suggestion
Shorebird can be used to update any and all Dart code in your app, including non-plugin dependencies.
```

docs/faq.md
```diff
@@ -111,6 +111,16 @@ Also Shorebird does not support changing native code (e.g. Java/Kotlin on
...
+Note that depending on how you distribute your apps, some store agreements
```
bryanoltman, line 117:
```suggestion
However, some store agreements
```

eseidel (approved):

docs/faq.md
felangel: Maybe we should say "include dart packages (not plugins)."

bryanoltman: "including packages that do not contain plugins"?

docs/faq.md
felangel: I think leaving it as is is more explicit personally 🤷

docs/faq.md
felangel: I went with "including pure dart packages" lmk what you think

PR #180, opened 2024-02-02
doc: add a doc on getting crash reporting symbols

eseidel:
We may just want to expose these as urls in the console instead?


felangel (approved):
LGTM

eseidel:
Links work now. 🎉

PR #190, opened 2024-03-12
Remove references of expo.dev

jonsamp:
## Status

READY

## Description

I saw that there are references to expo.dev on Shorebird's docs and we'd like to remove them, to avoid any confusion about how our service works. We heard from a few people drawing comparisons and it took us awhile to explain how we're different.

Overall, we are trying to center our updates product around developer workflows, and provide the most value there. Also, I want to make sure users don't think that Expo's updates allow more than we offer, like that we only allow updating an app's JS bundle for critical bug fixes in production and we don't support accessing any native APIs or functionality.

Shorebird is a great project and an awesome tool for Flutter developers, I'm just trying to make sure users don't get confused along the way.

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -16,8 +16,7 @@ Shorebird currently works on Android and iOS (alpha) and will eventually work
 everywhere Flutter works.
 
 "Code Push" is a reference to the name of a deploy feature used by the React
-Native community from [Microsoft](https://appcenter.ms) and
-[Expo](https://expo.dev), neither of which support Flutter.
+Native community from [Microsoft](https://appcenter.ms), neither of which support Flutter.
 
 ## Patching
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -87,7 +87,7 @@ your account.
 
 Code push services are widely used in the industry (all of the large
 apps I'm aware of use them) and there are multiple other code push services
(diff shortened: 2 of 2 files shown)


eseidel:
Hi John.  I'm not averse to changing our sites to avoid mentioning Expo, but I'm not sure I understand the problem being solved?  (Understanding will help me make sure that we're being helpful to you going forward.)

These two particular docs I'm sure have effectively zero page-rank relative to any documentation which Expo itself provides, given how young Shorebird is relative to Expo and how little usage we have.

eseidel:
Thanks for the outreach.  It's been a while since I've spoken with Charlie, but please don't hesitate to reach out to us if Shorebird is causing trouble for Expo.  You all have certainly been helpful to us in the past and we'd like to continue to be for you.  I'm going to close this for now based on my logic above, please let me know if I've misunderstood.  Thanks!

ide:
hi @eseidel - there are two things. One is that at a technical level Expo supports downloading only JS and no Java bytecode, dynamic libraries, or other artifacts that may add native capabilities. AFAIK (but could be mistaken) that is one difference from Shorebird's approach. We'd like to distinguish the two approaches and not put everything under one "code push" umbrella. The second thing is that CodePush is Microsoft's library and is unrelated to Expo.

eseidel:
Hi James.  Thanks for the follow-up.  I'll drop you an email.

PR #192, opened 2024-03-21
Add note about Xcode version differences causing native change warnings

bryanoltman:


eseidel (approved):
We may find we need to record this (and possibly other info) when recording the release. 🤷

bryanoltman:
@eseidel that's a great idea. This will probably be similar in concept to how we record force/asset/native changes when patching, although I'm thinking adding a DB column for each bit of metadata we may want to store was probably not the best approach. We could probably insert into a bigquery table on release/patch creation?

PR #204, opened 2024-04-08
feat: Don't repeat CI authentication docs

erickzanardo:
## Status

**READY**

## Description

This PR extracts the part of the docs that explain how to obtain a CI token in its own reusable file. Which then is used on both the github and codemagic guides. Removing the duplicated docs between both files.

This will make this section of the docs easier to scale as we add more content.

After this, I plan on adding to the `_ci_authentication.md` file about how the token can also be acquired from the console.

### DIFF
--- a/docs/ci/github.mdx
+++ b/docs/ci/github.mdx
@@ -4,6 +4,8 @@ title: 🐙 GitHub
 description: Integrate Shorebird into your GitHub workflow
 ---
 
+import CIAuthentication from './_ci_authentication.md';
+
 # GitHub Workflow Integration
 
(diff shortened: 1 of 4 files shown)


docs/ci/github.mdx
```diff
@@ -54,38 +56,7 @@ Currently `setup-shorebird` only supports the latest stable version of shorebird
 
 ## Authentication
```
felangel, line 57: Can we include the Authentication heading as part of the reusable CIAuthentication component?

felangel (approved):
LGTM just left one comment about moving the heading into the component as well 👍

docs/ci/github.mdx
erickzanardo: That was my first approach, but when doing so, the section would not show in the index 🙃

felangel: Ah that’s annoying haha

PR #234, opened 2024-06-03
fix: add missing redirects

eseidel:
These were reported by our Google Search Console, currently they're
404s in the new site.

Astro doesn't have landing pages for the directories, so redirected
to the android variants.


felangel (approved):
LGTM just need to run the formatter (`npm run format`)

PR #256, opened 2024-06-27
docs: adding Can the user delete patches from their device?

erickzanardo:
## Status

**READY**

## Description

Adds a new section `Can the user delete patches from their device?`

Which covers what happen if users delete the application data in their devices.

### DIFF
--- a/src/content/docs/faq.mdx
+++ b/src/content/docs/faq.mdx
@@ -248,9 +248,15 @@ system should you choose.
 
 ### Can the user delete patches from their device?
 
-Yes. Since the patches are store in the application data folder, if users
-delete the application data in their devices, the patches will deleted as well.
-This will make the application be sent to the original state of the release.
+Yes.
+
+Since patches are stored in the application data folder, if users
+go in the application settings inside their device system preferences and
+delete the app data, any stored patch will deleted as well.
+
+This should not break the application though. The application will be sent back to
+the original state of when the release was created and Shorebird will trigger
+an update to the latest patch next time the user open the app.
 
 ### How does this relate to Flutter Hot Reload?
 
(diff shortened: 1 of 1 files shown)


src/content/docs/faq.mdx
```diff
@@ -246,6 +246,12 @@ command works.
 [Uninstall](/uninstall) has details on how to remove Shorebird from your
 system should you choose.
 
+### Can the user delete patches from their device?
+
+Yes. Since the patches are store in the application data folder, if users
+delete the application data in their devices, the patches will deleted as well.
+This will make the application be sent to the original state of the release.
```
bryanoltman, line 253: I think I'd provide a little more context here (specifically, the steps the user needs to go through to do this). As it's written, this makes it sound like a very easy and obvious thing for a user to do.

erickzanardo: Tweaked a little bit, can you take another look please?

src/content/docs/faq.mdx
```diff
@@ -246,6 +246,18 @@ command works.
...
+
+Yes.
+
+Since patches are stored in the application data folder, if users
+go in the application settings inside their device system preferences and
+delete the app data, any stored patch will deleted as well.
+
+This should not break the application though. The application will be sent back to
+the original state of when the release was created and Shorebird will trigger
+an update to the latest patch next time the user open the app.
```
eseidel, line 259: I think this is Android-only.  I don't think there is any way to "clear application data" on iOS.

eseidel: Here is where we set the path on android (for your own amusement): https://github.com/shorebirdtech/engine/blob/a72fb61b0f4211313cabc1395e7ba062c5fa176a/shell/platform/android/flutter_main.cc#L130

erickzanardo: Fair enough! I checked on my iPhone and it seems that we really can't delete the data! TIL

I will mention that this is Android only

felangel:
Closing this for now

PR #262, opened 2024-07-03
Add "check in pubspec.lock" advice to troubleshooting

bryanoltman:
Adds advice to check in pubspec.lock in the native changes troubleshooting section.

### DIFF
--- a/src/content/docs/troubleshooting.mdx
+++ b/src/content/docs/troubleshooting.mdx
@@ -240,6 +240,11 @@ This can be caused by a number of things. The most common causes are:
    :::tip
    Add your `pubspec.lock` file to version control so you're always aware of
    updates to your dependencies.
+
+   Versioning your `pubspec.lock` file ensures changes to transitive
+   dependencies are explicit. Each time the dependencies change due to
+   `dart pub upgrade` or a change in `pubspec.yaml`, the difference will be
+   apparent in the lock file.
    :::
 
 2. A dependency/plugin produces a different output on every build. This can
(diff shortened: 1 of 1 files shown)


src/content/docs/troubleshooting.mdx
```diff
@@ -237,6 +237,11 @@ This can be caused by a number of things. The most common causes are:
    use caution when publishing patches that include changes to native code from
    plugins. In the worst case, these changes may cause your patched app to
    crash.**
+   :::tip
+   Add your `pubspec.lock` file to version control so you're always aware of
```
erickzanardo, line 241: Dart has a great guide about the lock file, should we link it too?

https://dart.dev/guides/libraries/private-files#pubspec-lock

erickzanardo (approved):
One suggestion, but looks great already!

src/content/docs/troubleshooting.mdx
bryanoltman: If there were a Flutter-specific guide, I'd say yes. The dart guide is a little less clear/direct than I'd like (basically, I just want to tell users "Yes, check it in" rather than "don't check it in _unless_ your package is an application package" and then having the users wonder what an application package is and whether their Flutter app is one of those, etc.). I'll copy some of the text from those docs over.

erickzanardo: Sounds good, my suggestion was more in the lines that this guide corroborate for what we are suggesting the users.

But no strong opinions, the content is already good!

PR #268, opened 2024-07-16
Add flags to native/asset change troubleshooting sections

bryanoltman:
Updates troubleshooting docs to include the flags required to bypass native/asset change warnings when patching.

This also removes what was supposed to be a custom header slug that was rendering inline: 
Screenshot 2024-07-16 at 2 31 09 PM

### DIFF
--- a/src/content/docs/troubleshooting.mdx
+++ b/src/content/docs/troubleshooting.mdx
@@ -147,7 +147,7 @@ version and build numbers in order to target patches at specific releases. You
 can fix this problem by either setting `manageAppVersionAndBuildNumber` to false
 or removing the value from your export options .plist file.
 
-## I see a `Your app contains asset changes` warning when running `shorebird patch` (#asset-changes)
+## I see a `Your app contains asset changes` warning when running `shorebird patch`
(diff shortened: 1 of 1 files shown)


src/content/docs/troubleshooting.mdx
```diff
@@ -147,7 +147,7 @@ version and build numbers in order to target patches at specific releases. You
 can fix this problem by either setting `manageAppVersionAndBuildNumber` to false
 or removing the value from your export options .plist file.
 
-## I see a `Your app contains asset changes` warning when running `shorebird patch` (#asset-changes)
+## I see a `Your app contains asset changes` warning when running `shorebird patch`
```
eseidel, line 150: Does the #asset-changes thing do anything?  Was that the perma-link name?

bryanoltman: I think this worked before the migration to astro to make the link to this page be troubleshooting#asset-changes. I couldn't find any links in our CLI or our docs site pointing to this header.

eseidel (approved):

PR #275, opened 2024-07-30
Update fastlane guide to reflect plugin update

bryanoltman:
Our new version of the fastlane plugin autodetects the ipa path, so it is no longer necessary to explicitly pass this value.

### DIFF
--- a/src/content/docs/ci/fastlane.mdx
+++ b/src/content/docs/ci/fastlane.mdx
@@ -11,8 +11,9 @@ releasing and patching.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


eseidel (approved):

eseidel:
I presume our plugin still (optionally) takes an ipa parameter to not break existing users?

bryanoltman:
> I presume our plugin still (optionally) takes an ipa parameter to not break existing users?

`upload_to_testflight` isn't ours. The update has our plugin pass the IPA path via an env variable instead of making the user specify it.

PR #286, opened 2024-09-03
feat: Update FAQ about sizes on disk

eseidel:


src/content/docs/faq.mdx
```diff
@@ -250,11 +250,24 @@ although the base of that path is provided by the Android system and can change
...
+### What do the Shorebird tools store on disk?
+
 The Shorebird command line tools (e.g. `shorebird patch`) are installed on disk
 in `$HOME/.shorebird`, including bringing a copy of Flutter and Dart as well as
 the Shorebird release tools themselves. This is similar to how the `flutter`
 command works.
 
+These copies of Flutter can be many 100s of megabytes each. Currently Shorebird
+will keep an unbounded number of local copies of Flutter in
+`$HOME/.shorebird/bin/cache/`.
```
felangel, line 269: Maybe we should add you can clear cached artifacts to free up disk space via `shorebird cache clear`?

felangel (approved):

src/content/docs/faq.mdx
```diff
@@ -250,11 +250,24 @@ although the base of that path is provided by the Android system and can change
 dynamically at runtime. On iOS devices, data is stored under
 `Library/Application Support/shorebird`.
 
+Shorebird stores only a few files, one is `state.json` which contains the state
+information for the updater. When patches are installed there is also a "slots"
+directory containing up to two directories (0, 1) for each potential downloaded
+patch. These will contain the inflated copies of the patches, which will be
```
bryanoltman, line 256: This bit about slots is no longer true. Patches are now stored in their own directory, and we should never have more than two on disk at the same time. Once we've successfully booted from patch N+1, patch N's artifacts are deleted.

bryanoltman (approved):

src/content/docs/faq.mdx
eseidel: Fixed.

PR #299, opened 2024-10-09
docs: Update our teams documentation

AhmedLSayed9:
Remove `Adding an Admin` request from docs as it can currently be done in the console.


eseidel:
Thanks!  We just shipped "organizations" support today and @felangel just re-wrote this page in https://github.com/shorebirdtech/docs/pull/300.  Thanks for the PR all the same!

PR #308, opened 2024-11-13
update patch docs to reflect track

Mastersam07:
## Status

**READY**

## Description

In older shorebird versions, we stage a patch by doing:
```sh
shorebird patch android --staging
shorebird patch ios --staging
```

Currently, --staging flag is deprecated and the updated way to stage a patch is by doing:
```sh
shorebird patch android --track=staging
shorebird patch ios --track=staging
```

This PR aims to update the doc reflecting this change.


bryanoltman (approved):
Thank you!

PR #311, opened 2024-11-13
docs: Add a page on billing.

eseidel:
Eventually we should pull in content from https://shorebird.dev/blog/simplified-pricing/

### DIFF
--- a/src/content/docs/billing.mdx
+++ b/src/content/docs/billing.mdx
@@ -1,0 +1,35 @@
+---
+title: Billing
+description: How Shorebird handles Billing
+---
+
+Shorebird uses Stripe for all billing. We accept payment in any method
+that Stripe accepts.
+
+Annual billing, invoice billing and other payment methods are also available to
+customers purchasing over 1M patches per month, or custom enterprise plans. See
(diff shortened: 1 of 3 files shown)


src/content/docs/billing.mdx
```diff
@@ -0,0 +1,35 @@
+---
+title: Billing
+description: How Shorebird handles Billing
+---
+
+Shorebird uses Stripe for all billing. We accept payment in any method
```
felangel, line 6: Seems very closely related to pricing. I wonder if we shouldn't at least link to billing from pricing/vice-versa or even consider consolidating this into the pricing page 🤷

felangel (approved):

src/content/docs/billing.mdx
eseidel: I agree, they are related.

bryanoltman (approved):

src/content/docs/billing.mdx
eseidel: We can clean them up further.  Landing this for now.

PR #315, opened 2024-12-02
docs: add a section on automatic rollback

eseidel:
I didn't find us documenting this anywhere, so I wrote some docs.

Not sure if this is the best place, but we can always move it later.

### DIFF
--- a/src/content/docs/code-push/rollback.mdx
+++ b/src/content/docs/code-push/rollback.mdx
@@ -50,8 +50,8 @@ have no remaining patch installs, the app will revert to the base release.
 
 ## Patch Integrity and Automatic Rollback
 
-Shorebird includes safety features to prevent needing to issue the manual roll
-back of a patch.
+Shorebird includes safety features to prevent needing to issue the manual
+rollback of a patch.
 
 `shorebird patch` also generates a hash of your patch file which is stored
 separately from the patch and checked every time your patch is installed, or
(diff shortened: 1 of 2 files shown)


src/content/docs/code-push/rollback.mdx
```diff
@@ -47,3 +47,36 @@ If you app needs to download an older patch as part of a rollback,
 installing that patch will count against your monthly patch installs. If you
 have no remaining patch installs, the app will revert to the base release.
 :::
+
+## Patch Integrity and Automatic Rollback
+
+Shorebird includes safety features to prevent needing to issue the manual roll
+back of a patch.
```
bryanoltman, line 54:
```suggestion
Shorebird includes safety features to prevent needing to issue the manual
rollback of a patch.
```

src/content/docs/code-push/rollback.mdx
```diff
@@ -47,3 +47,36 @@ If you app needs to download an older patch as part of a rollback,
...
+
+`shorebird patch` also generates a hash of your patch file which is stored
+separately from the patch and checked every time your patch is installed, or
+booted from on the user's device.
+
+Customers are also encouraged to use Shorebird's [patch-signing
```
bryanoltman, line 60: Not sure if necessary, but I think we still need to support this for hybrid app targets  https://github.com/shorebirdtech/shorebird/issues/2162

bryanoltman (approved):

src/content/docs/code-push/rollback.mdx
eseidel: That's fine, we'll add it when requested.

bryanoltman: My suggestion was more "should we mention it here so someone who reads this and tries isn't confused"

eseidel: Nah, I think that goes in the signing docs themselves.  We also just generally need a "make our system awesome for hybrid apps" pass over our whole product, that I think comes later as we're more established.

PR #316, opened 2024-12-02
docs: do I still need to submit to the stores?

eseidel:
We've gotten this question a few times, figure it should go in the FAQ.

### DIFF
--- a/src/content/docs/faq.mdx
+++ b/src/content/docs/faq.mdx
@@ -42,6 +42,8 @@ The following URLs are used by Shorebird:
 - https://storage.googleapis.com -- used by the `shorebird` command line tool to
   upload and download release and patch artifacts, and by the Shorebird updater
   on user's devices to download the patches.
+- https://cdn.shorebird.cloud/ -- used by the Shorebird updater when downloading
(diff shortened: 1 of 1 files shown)


src/content/docs/faq.mdx
```diff
@@ -475,6 +477,23 @@ If you'd like to segment availability of Shorebird patches, there are 4 potentia
 
 Yes. Apps built with Shorebird will continue to function normally (as if they had been built without Shorebird), including those that have had patches installed.
 
+### Now that I'm using Shorebird, do I still need to send my app to the stores?
```
bryanoltman, line 480: Should this be "will shorebird handle all distribution for me"? Not all of our users distribute through the stores.

eseidel: We actually already have that question: https://docs.shorebird.dev/faq/#does-shorebird-submit-to-the-stores-for-me. I can link to one from the other?

eseidel: I've re-sorted the answers.

bryanoltman (approved):

PR #336, opened 2025-02-05
docs: add Linux

felangel:
## Status

**READY**

## Description

- docs: add Linux


eseidel (approved):
I'd probably remove the (beta) everywhere, since I don't expect that to be in place very long and we already warn at the CLI level.

PR #372, opened 2025-05-13
feat: add a new doc about iOS performance

eseidel:


src/content/docs/code-push/performance.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+In that example, `isEven` and `foo` didn't change. So you would expect Shorebird
+should be able to run both of those on the CPU (not have to use an interpreter).
+However this is unfortunately not the case. In the "before" case, because isEven
+was never called with nullable values Dart optimized away all of the null checks
+in the compiled `isEven` code. (That's a good thing.) Unfortunately that means
+that when the new code now used isEven in a nullable context the compiled
+contents of "isEven" change, to include null checks. Since the contents of
+isEven changed, Shorebird will conservatively assume that the behavior of all
+callers of isEven might have changed as well and we will not use the CPU
```
felangel, line 81:
```suggestion
callers of `isEven` might have changed as well and we will not use the CPU
```

src/content/docs/code-push/performance.mdx
```diff
@@ -0,0 +1,103 @@
...
+  print(isEven(x));
+}
+```
+
+In that example, `isEven` and `foo` didn't change. So you would expect Shorebird
+should be able to run both of those on the CPU (not have to use an interpreter).
+However this is unfortunately not the case. In the "before" case, because isEven
+was never called with nullable values Dart optimized away all of the null checks
+in the compiled `isEven` code. (That's a good thing.) Unfortunately that means
+that when the new code now used isEven in a nullable context the compiled
```
felangel, line 78:
```suggestion
that when the new code now used `isEven` in a nullable context the compiled
```

src/content/docs/code-push/performance.mdx
```diff
@@ -0,0 +1,103 @@
...
+bool isEven<T extends int?>(T value) => value?.isEven ?? false;
+
+void foo(int x) {
+  print(isEven(x));
+}
+
+void bar(int? x) {
+  print(isEven(x));
+}
+```
```
felangel, line 71:
```suggestion
```diff
bool isEven<T extends int?>(T value) => value?.isEven ?? false;

void foo(int x) {
  print(isEven(x));
}

+ void bar(int? x) {
+   print(isEven(x));
+ }
```
```

felangel (approved):

PR #380, opened 2025-05-30
Adjust "testing patches" to remove reference to fixed tracks

bryanoltman:
Part of https://github.com/shorebirdtech/updater/issues/255


felangel (approved):
LGTM we should also probably update screenshots of promoting to beta/staging etc in the development workflow guide.

PR #426, opened 2025-09-05
Feat/new ci install steps

tomarra:
## Status

**READY**

## Description

Updating the CI Setup page with the new setup flow including screenshots


tomarra:
> They can select "all" repos and it will work just fine.

I think this was more to also save us some run costs for when we get triggered on non Dart repos and it essentially becomes a no-op. @felangel @bryanoltman can confirm that though.

felangel (approved):

PR #443, opened 2025-11-04
Make patch signing less beta-y

bryanoltman:
Removes negative language about readiness of patch signing, encourages companies to reach out if the current solution doesn't meet their needs.

### DIFF
--- a/src/content/docs/code-push/guides/patch-signing.mdx
+++ b/src/content/docs/code-push/guides/patch-signing.mdx
@@ -65,6 +65,14 @@ stored securely and kept secret. While the private key is not itself sufficient
 to make an update to your application (someone would also need access to your
 Shorebird credentials), it should not be checked into public source control.
 
+:::note
+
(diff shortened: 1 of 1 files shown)


src/content/docs/code-push/guides/patch-signing.mdx
```diff
@@ -7,19 +7,6 @@ sidebar:
 
 {/* cspell:words outform pubout */}
 
-:::note
-
-We've shipped this version of patch signing to unblock certain customer
-evaluations. While we're confident that this feature adds a useful layer of
-security, it does not yet have all the expected integrations with cloud signing,
-keychains, etc. that a final product would expect to have. We also do not yet
-manage signing keys for you or support many key formats or signature types.
```
tomarra: Wondering if the fact that we don't manage your key's for you is worth keeping around so people don't make the assumption that we do like the stores. Thoughts?

bryanoltman: Updated

PR #444, opened 2025-11-05
fix: update note about patch signing

tomarra:
## Status

**READY**

## Description

Remove a note about us not supporting patch signing and instead point readers to our patch signing page if needed.


src/content/docs/code-push/system-architecture.mdx
```diff
@@ -144,8 +144,9 @@ The hash is not meant to be a security feature, but rather a way to detect
 errors in the patch diff. A common error we see is that developers
 `shorebird release` with one source and then actually build and release a
 different binary, resulting then in invalid patches. This hash helps detect such
-errors. We currently do not sign patches, but plan to add that capability in the
-near future.
+errors. For added security we also provide a way for you to
```
bryanoltman:
```suggestion
errors. For added security, we also provide a way for you to
```

PR #488, opened 2026-02-05
feat: clarify the release delete flow

tomarra:
## Status

**READY**

## Description

Update the Delete Release section to include the fact that you can delete releases by platform.

CleanShot 2026-02-05 at 14 31 53@2x


easymac:
Looks like a real spell check problem

Then looks great!

PR #552, opened 2026-06-08
feat: add roadmap page

tomarra:
## Status

**READY** 

## Description

Adding in a public roadmap page that is generated based on our Linear project list. This pulls any project that is labeled with the "Public Roadmap" label. This is regenerated on a nightly basis to ensure data is fresh and its done via Actions so the Linear API key is secured in GitHub Secrets and not exposed via a call on the client. There is also an option to just manually redeploy at any time as well via workflow dispatch.

This shows items going back one quarter and then forward looking as well as items that don't have a timeline targeted. This means that when we get to Q3 the Q1 items will magically be removed from this page in order to ensure it doesn't continue to grow indefinitely. 

Note: Linear API keys are scoped to a user instead of an account. If we were to do the account level it would have to be a OAuth app which is way overkill here. This means the stored API key is currently tied to my account https://linear.app/docs/api-and-webhooks#api-keys

**Preview Screenshot** 
CleanShot 2026-06-08 at 11 54 02@2x


src/pages/roadmap.astro
```diff
@@ -0,0 +1,285 @@
...
+              name
+              description
+              state
+              targetDate
+              priority
+            }
+          }
+        }`,
+      }),
+    });
```
nickshorebird, line 85: Consider adding a timeout here or to this workflow/file as a whole to hedge against a hung process with no default timeouts.

nickshorebird (approved):
Consider having a workflow create a PR and tagging a set of users to approve if changes are seen. This would hedge against someone accidentally having PII become public via a tag miss set. This would also separate the the actual doc updates from the roadmap doc update logic and could run on longer set schedule.

src/pages/roadmap.astro
tomarra: Good call. Added.

tomarra:
> Consider having a workflow create a PR and tagging a set of users to approve if changes are seen. This would hedge against someone accidentally having PII become public via a tag miss set. This would also separate the the actual doc updates from the roadmap doc update logic and could run on longer set schedule.

Since we are limiting what we are grabbing out of Linear I'm less concerned about the PII side of things (also there shouldn't be any PII in Linear in general). The PR thing is an interesting one I haven't thought about. Probably wouldn't be hard to do if we deem that to be useful.

nickshorebird:
> > Consider having a workflow create a PR and tagging a set of users to approve if changes are seen. This would hedge against someone accidentally having PII become public via a tag miss set. This would also separate the the actual doc updates from the roadmap doc update logic and could run on longer set schedule.
> 
> Since we are limiting what we are grabbing out of Linear I'm less concerned about the PII side of things (also there shouldn't be any PII in Linear in general). The PR thing is an interesting one I haven't thought about. Probably wouldn't be hard to do if we deem that to be useful.

Sorry meant confidential not PII.

tomarra:
Holding this one until we get the thumbs up from @eseidel

eseidel:
I'm in favor of the idea.  My only concern reading this is that it's non-obvious what fields in Linear now become public-by-default vs. private after this were to land.

eseidel:
I see, re-reading, it's tag-based.  Specifically the "public roadmap" tag makes it public.

eseidel (approved):

PR #554, opened 2026-06-09
fix: add rermak parser back in to fix tables

tomarra:
## Status

**READY**

## Description

Starlight calls `mdx({ optimize: true })` with no `gfm` option, and Astro v6 removed the `gfm: true` default from the markdown config schema. So the MDX integration's `if (mdxOptions.gfm)` check was always false, and remark-gfm was never added to the MDX pipeline.

The fix explicitly adds remarkGfm to markdown.remarkPlugins, which markdownConfigToMdxOptions passes directly into the MDX pipeline bypassing the broken gfm flag entirely.


AbhishekDoshi26 (approved):
LGTM!!!

PR #576, opened 2026-07-02
docs: add style checks for headings and content guidelines

tomarra:
## Status

**READY** (pending #583 merging first — see below)

## Description

Adds automated checks for a first draft of content guidelines from marketing, using [Vale](https://vale.sh) (a prose style linter) plus a small companion script for things Vale can't see.

**What's checked now (`npm run lint:content`):**
- Headings (`##`, `###`, ...) use sentence case, not Title Case (`Shorebird.Headings`) — **blocking**.
- No exclamation points in prose (`Shorebird.Exclamation`), ignoring code spans/blocks and markdown image syntax — **blocking**.
- `<TabItem label="...">` and `<LinkButton>` text follow the same sentence-case / upper-case rules, via `scripts/lint-component-labels.mjs` — Vale only parses markdown prose, so it can't see MDX/JSX attributes or isolate component boundaries. Local-only, not wired into CI.
- "Shorebird", "Flutter", and "Code Push" are always capitalized correctly anywhere in prose, not just headings (`Vale.Terms`, driven by a Shorebird `Vocab`) — **non-blocking** (see below for why).
- Heuristic check for first-person pronouns (I/we/our/us) in body paragraphs, nudging toward second person (`Shorebird.SecondPerson`) — **non-blocking**, content work not started yet.

**CI is now blocking** (`fail_on_error: true`) for `Shorebird.Headings` and `Shorebird.Exclamation`, which are fully clean after [#577](https://github.com/shorebirdtech/docs/pull/577), [#578](https://github.com/shorebirdtech/docs/pull/578), [#580](https://github.com/shorebirdtech/docs/pull/580), and [#583](https://github.com/shorebirdtech/docs/pull/583) merge. `Vale.Terms` and `Shorebird.SecondPerson` stay at warning severity (Vale's exit code, and therefore `fail_on_error`, only responds to error-severity alerts) — still reported on every PR, just don't block.

**⚠️ Merge-order dependency: do not merge this PR before [#583](https://github.com/shorebirdtech/docs/pull/583) (heading sentence-case fixes) and [#584](https://github.com/shorebirdtech/docs/pull/584) (a small regression fix) land.** This branch's own CI will show red until then, since its content still has the original ~162 Title Case headings. I verified the full combination (this branch + `main` + #583 + #584) passes cleanly with `fail_on_error: true` before pushing this.

[...]

PR #583, opened 2026-07-02
docs: use sentence case for headers

tomarra:
## Status

**READY**

## Description

Per the marketing content guidelines (see [#576](https://github.com/shorebirdtech/docs/pull/576)), headers should use sentence case, not Title Case. Fixes 169 headings across 44 files.

This one took more digging than the others — Vale's capitalization rule has some real quirks I only found by testing directly against it rather than guessing. Details below since they matter for review.

**Left untouched — compound product/brand names** where fixing the flagged word in isolation makes the phrase read worse than leaving it alone (e.g. "Azure key Vault", "app Store connect"):
- "Azure Key Vault", "GCP Cloud KMS", "Create a release in App Store Connect"
- "How does this relate to Firebase Remote Config or Launch Darkly?"
- "How does this relate to Flutter Hot Reload?" (Flutter's own name for the feature)

**Left untouched — three "Flutter vs. X" headings:** Vale's capitalization rule flags any heading containing "vs" or "vs." regardless of what surrounds it (verified with isolated test cases — even a heading of just "Testing vs here" gets flagged). This isn't configurable via the exceptions list; it's baked into the rule. The content itself (`Flutter vs. the competition`, `Flutter vs. native development`) is already correctly cased; `Flutter vs. React Native` is left as-is since "React Native" is itself a proper noun.

**Left untouched — four "code push" FAQ headings** (e.g. "Does code push require the internet to work?"): enabling the `Vocab`-driven proper-noun rule (for the Shorebird/Flutter/Code Push guideline) has a side effect — it makes Vale's *capitalization* rule also flag any heading containing a mis-cased vocab term, even when the heading has no actual sentence-case violation. These four headings get fixed by [#579](https://github.com/shorebirdtech/docs/pull/579) (capitalizing "code push" → "Code Push"), which resolves this side effect too. No separate edit needed here.

**Reverted to Title Case — a 3-item numbered list** ("1. The Status Enum", "2. The Single State Class", "3. The Events"): oddly, Vale stops flagging #2 once lowercased but keeps flagging #1 and #3 despite identical structure (confirmed this is a real Vale inconsistency, not a mistake on my end). Rather than leave the list looking inconsistent, I reverted all three to their original Title Case.

**New proper nouns discovered and added to the exceptions list** (not previously recognized): Stripe, Fastfile, BuildContext, RenderObject, Apple, 1Password, AOT, GuardSquare, LTS, ExportOptions.plist, Skia, Impeller. Also fixed two headings using lowercase "fastlane" where the tool name should be capitalized ("Using Fastlane on CI/locally").

**Follow-up needed:** those new exceptions need to land in `.vale/styles/Shorebird/Headings.yml` on the tooling PR (#576) too — otherwise they'll show up as false positives again once both PRs merge. I'll push that update to #576 separately.

This is one of several small content PRs split out of #576 so the `Shorebird.Headings` check can eventually be made blocking without a giant, hard-to-review PR.

## Test plan

- [x] `npm run build` succeeds, no broken links
- [x] `cspell` passes on all touched files
- [x] Verified locally with the (not-yet-merged) Vale config from #576 (plus the new exceptions above) — only the 15 intentionally-left findings above remain


src/content/docs/code-push/guides/hybrid-apps/ios.mdx
dawn-ducky, line 1: Framework Search Paths appears to be a proper name and should be capitalized.

src/content/docs/flutter-concepts/architecture-trees.mdx
```diff
@@ -33,7 +33,7 @@ graph TD
     style R fill:#fff3e0,stroke:#ff9800,stroke-width:2px,color:#000
 ```
 
-## 1. The Widget Tree: Your Blueprints
```
dawn-ducky, line 36: "The" should be capitalized

src/content/docs/flutter-concepts/architecture-trees.mdx
```diff
@@ -50,7 +50,7 @@ When you call `setState()`, you aren’t directly telling the screen to redraw.
 You are just telling Flutter to throw away the old stale objects and create new
 ones from the blueprints you created.
 
-## 2. The Element Tree: The State Manager
```
dawn-ducky, line 53: "The" should be capitalized

src/content/docs/flutter-concepts/architecture-trees.mdx
```diff
@@ -77,7 +77,7 @@ This is the secret to Flutter's speed! The Element Tree manages the lifecycle
 and acts as the smart middleman that decides when actual heavy lifting needs to
 be done.
 
-## 3. The RenderObject Tree: The Workhorse
+## 3. the RenderObject tree: the workhorse
```
dawn-ducky: "The" should be capitalized

PR #590, opened 2026-07-10
docs: add comprehensive guide to Flutter animations

AbhishekDoshi26:
## Description
This PR adds a highly technical, objective reference guide to Flutter animations under the `flutter-concepts` section.

The document strips away conversational fluff and focuses purely on the underlying architecture and APIs required for high-performance motion in Flutter.

### Highlights:
- Added `animations.mdx` (sidebar order 6).
- **Implicit Animations**: Covers `ImplicitlyAnimatedWidget` internals and non-linear interpolation (Curves).
- **Explicit Animations**: Details `AnimationController`, `TickerProvider` (vsync), `CurvedAnimation`, and `Tween` mapping. Includes a Mermaid flow diagram visualizing the architecture.
- **Physics-based Simulations**: Provides examples for `SpringSimulation` and `controller.animateWith()`.
- **Performance Considerations**: Outlines best practices for avoiding layout passes by animating opacity and transform properties directly on the GPU.
- **Linting**: The document strictly complies with all Vale rules (no first-person pronouns, sentence case headings, no exclamation points).

### DIFF
--- a/src/content/docs/flutter-concepts/animations.mdx
+++ b/src/content/docs/flutter-concepts/animations.mdx
@@ -1,0 +1,247 @@
+---
+title: 'Flutter Animations'
+description:
+  Reference guide for implementing implicit, explicit, and physics-based
+  animations in Flutter.
+sidebar:
+  order: 6
+---
+
+The Flutter animation framework provides a layered architecture for controlling
+motion. At its core, the framework categorizes animations into implicit
+transitions, explicit controller-driven animations, and physics-based
+simulations.
+
+## Implicit animations
+
+Implicit animations automatically interpolate between a current value and a new
+target value over a specified duration. The framework calculates intermediate
+frames when a property changes and the widget rebuilds.
+
+`AnimatedContainer` is a versatile implicit widget that can animate changes to
+size, color, padding, and alignment simultaneously:
+
+```dart
+AnimatedContainer(
+  duration: const Duration(milliseconds: 300),
+  curve: Curves.easeInOut,
+  width: isExpanded ? 200 : 100,
+  decoration: BoxDecoration(
+    color: isExpanded ? Colors.blue : Colors.red,
+    borderRadius: BorderRadius.circular(isExpanded ? 16.0 : 8.0),
+  ),
+  child: const Text('Hello'),
+)
+```
+
+The underlying architecture relies on `ImplicitlyAnimatedWidget`. When the
+widget's configuration changes (detected in `didUpdateWidget`), the state reuses
+its internal `AnimationController`, rebases the tween's `begin` to the current
+animation value, and restarts the transition toward the new target. This is why
+implicit animations transition smoothly when interrupted mid-flight.
+
+### Non-linear interpolation (curves)
+
+By default, animations transition linearly. Applying a `Curve` modifies the
+interpolation rate. The framework provides standard curves via the `Curves`
+class, such as `Curves.easeIn`, `Curves.bounceOut`, and `Curves.elasticInOut`.
+You can also define custom mathematical curves by subclassing `Curve` and
+overriding the `transformInternal` method.
+
+## Explicit animations
+
+Explicit animations require manual lifecycle management using an
+`AnimationController`. This approach is necessary for looping animations,
+sequenced transitions, and animations triggered by gesture events.
+
+```mermaid
+graph TD
+    T["TickerProvider<br>(vsync)"] -->|Tick per frame| C["AnimationController<br>(0.0 to 1.0)"]
+    C -->|Linear Value| A["CurvedAnimation<br>(Non-linear timing)"]
+    A -->|Eased Value| TW["Tween<br>(Maps to target data type)"]
+    TW -->|Output Value| W["AnimatedBuilder<br>(Rebuilds UI)"]
+
+    style T fill:#e1f5fe,stroke:#03a9f4,stroke-width:2px,color:#000
+    style C fill:#e8f5e9,stroke:#4caf50,stroke-width:2px,color:#000
+    style A fill:#fff3e0,stroke:#ff9800,stroke-width:2px,color:#000
+    style TW fill:#fce4ec,stroke:#e91e63,stroke-width:2px,color:#000
+    style W fill:#f3e5f5,stroke:#9c27b0,stroke-width:2px,color:#000
+```
+
+### Animation controllers and vsync
+
+An `AnimationController` generates a new value whenever the hardware is ready
+for a new frame. To prevent off-screen computations and synchronize with the
+device refresh rate, the controller requires a `TickerProvider` (passed via the
+`vsync` argument).
+
+```dart
+class _MyWidgetState extends State<MyWidget> with SingleTickerProviderStateMixin {
+  late final AnimationController _controller;
+
+  @override
+  void initState() {
+    super.initState();
+    _controller = AnimationController(
+      duration: const Duration(seconds: 2),
+      vsync: this,
+    );
+  }
+
+  @override
+  void dispose() {
+    // Not calling dispose() leaks the Ticker. In debug mode, the mixin's
+    // dispose() throws "was disposed with an active Ticker" if the controller
+    // is still animating when the State is removed from the tree.
+    _controller.dispose();
+    super.dispose();
+  }
+}
+```
+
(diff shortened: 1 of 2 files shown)


src/content/docs/flutter-concepts/animations.mdx
```diff
@@ -0,0 +1,246 @@
...
+    // Failing to call dispose() causes a 'setState() called after dispose()'
+    // exception and leaks the Ticker.
+    _controller.dispose();
+    super.dispose();
+  }
+}
+```
+
+When managing multiple controllers in a single widget, use
+`MultiTickerProviderStateMixin` instead of `SingleTickerProviderStateMixin`.
```
bdero: `MultiTickerProviderStateMixin` doesn't exist in Flutter. The multi-ticker variant is just `TickerProviderStateMixin`. Needs updating here and in the code block below, which won't compile as written.

src/content/docs/flutter-concepts/animations.mdx
```diff
@@ -0,0 +1,246 @@
...
+);
+
+final SpringSimulation simulation = SpringSimulation(
+  spring,
+  0.0, // start position
+  100.0, // end position
+  0.0, // initial velocity
+);
+
+_controller.animateWith(simulation);
```
bdero: `AnimationController` clamps simulation values to `[lowerBound, upperBound]`, which defaults to `[0.0, 1.0]`. Driving this spring from 0.0 to 100.0 with the controller from the earlier example clamps at 1.0 almost immediately, so the spring motion is lost. The `animateWith` doc comment recommends `AnimationController.unbounded` for exactly this case. Suggest constructing the controller with `AnimationController.unbounded(vsync: this)` here, or animating over a 0.0 to 1.0 range instead.

src/content/docs/flutter-concepts/animations.mdx
```diff
@@ -0,0 +1,246 @@
...
+    _controller = AnimationController(
+      duration: const Duration(seconds: 2),
+      vsync: this,
+    );
+  }
+
+  @override
+  void dispose() {
+    // Failing to call dispose() causes a 'setState() called after dispose()'
+    // exception and leaks the Ticker.
```
bdero: Not disposing the controller doesn't throw `setState() called after dispose()`. That error comes from calling `setState` on an unmounted `State`, which is a different bug. What actually happens is a leaked `Ticker`, and in debug mode the mixin's `dispose` throws "was disposed with an active Ticker" if the controller is still animating. Suggest rewording the comment to describe that instead.

src/content/docs/flutter-concepts/animations.mdx
```diff
@@ -0,0 +1,246 @@
...
+    borderRadius: BorderRadius.circular(isExpanded ? 16.0 : 8.0),
+  ),
+  child: const Text('Hello'),
+)
+```
+
+The underlying architecture relies on `ImplicitlyAnimatedWidget`. When the
+framework detects a change in the widget's configuration (via
+`didUpdateWidget`), it creates an internal `AnimationController` and begins a
+transition from the old values to the new values.
```
bdero: Two inaccuracies here. `ImplicitlyAnimatedWidgetState` creates a single controller once for the life of the `State`, not a new one per change. And on a target change it doesn't animate from the old values, it rebases the tween's `begin` to the animation's current value and reruns `controller.forward(from: 0.0)`. That's why implicit animations retarget smoothly when interrupted mid-flight. Suggest something like "When the widget's configuration changes (detected in `didUpdateWidget`), the state reuses its internal `AnimationController`, rebases the tween's `begin` to the current animation value, and restarts the transition toward the new target."

src/content/docs/flutter-concepts/animations.mdx
```diff
@@ -0,0 +1,246 @@
...
+class _MyWidgetState extends State<MyWidget> with MultiTickerProviderStateMixin {
+  late final AnimationController _controllerA;
+  late final AnimationController _controllerB;
+}
+```
+
+### Tweens and mapping
+
+A controller typically outputs values between `0.0` and `1.0`. A `Tween` maps
+this unit interval to a required data type (such as `Color`, `Offset`, or
```
bdero: Listing `Color` here is misleading. The base `Tween` interpolates via `begin + (end - begin) * t`, and `Color` doesn't define those operators, so `Tween<Color>` throws at runtime. Colors need the specialized `ColorTween`. Worth a sentence noting that types without arithmetic operators use dedicated subclasses such as `ColorTween` and `RectTween`.

src/content/docs/flutter-concepts/animations.mdx
```diff
@@ -0,0 +1,246 @@
...
+  changes.
+
+## Performance considerations
+
+Modifying properties that affect layout (such as `width`, `height`, or
+`padding`) forces the framework to recalculate geometry on every frame, which
+can cause dropped frames.
+
+To optimize performance, animate properties that are composited directly on the
+GPU without altering layout constraints:
```
bdero: Flutter doesn't have a CSS-style set of GPU-composited properties, so this framing doesn't quite carry over. The actual win with `FadeTransition`, `ScaleTransition`, etc. is that they mutate render object properties at paint time, skipping the build and layout phases entirely. The frame still gets repainted. Suggest rewording to something like "animate properties that are applied during the paint phase, avoiding rebuilds and relayout".

src/content/docs/flutter-concepts/animations.mdx
```diff
@@ -0,0 +1,246 @@
...
+  `SlideTransition`.
+
+When using `AnimatedBuilder`, always pass static child widgets via the `child`
+parameter. This prevents the framework from recreating the child subtree during
+every animation tick.
+
+For complex subtrees that animate independently from the rest of the screen,
+wrap them in a `RepaintBoundary`. This tells the compositor to rasterize that
+subtree into a separate layer, so only that layer is repainted on each frame
+rather than the entire scene.
```
bdero: `RepaintBoundary` creates a separate layer so repaints of the animated subtree don't force the surrounding tree to repaint (and vice versa). Whether the engine additionally caches the layer as a raster texture is a separate raster-cache heuristic, not something the widget guarantees. Suggest "This isolates the subtree into its own layer, so its repaints don't propagate to the rest of the tree."

src/content/docs/flutter-concepts/animations.mdx
```diff
@@ -0,0 +1,246 @@
...
+
+```dart
+RepaintBoundary(
+  child: AnimatedBuilder(
+    animation: _controller,
+    builder: (BuildContext context, Widget? child) {
+      return Transform.rotate(
+        angle: _controller.value * 2 * 3.14159,
```
bdero: Minor nit. Use `math.pi` from `dart:math` instead of a hardcoded 3.14159. Alternatively, this is exactly what `RotationTransition(turns: _controller)` is for, which would also reinforce the transition-widgets advice above.

src/content/docs/flutter-concepts/animations.mdx
```diff
@@ -0,0 +1,246 @@
...
+_controller.animateWith(simulation);
+```
+
+## Third-party vector animations
+
+For complex vector graphics and state-machine animations, use third-party
+formats rather than the built-in widget framework:
+
+- **[Lottie](https://lottiefiles.com/)**: Uses animations exported as JSON from
+  Adobe After Effects. Best for static, non-interactive visual assets.
```
bdero: Minor nit. "Static" is an odd word for an animation format. Suggest "predetermined, timeline-based animations" to contrast with Rive's state machines.

src/content/docs/flutter-concepts/animations.mdx
```diff
@@ -0,0 +1,246 @@
...
+      height: sizeAnimation.value,
+      color: Colors.blue,
+      child: child, // passed through; not rebuilt on each tick
+    );
+  },
+  child: const FlutterLogo(),
+)
+```
+
+Alternatively, `TweenAnimationBuilder` provides a declarative API for one-shot
```
bdero: Minor nit. Not strictly one-shot. `TweenAnimationBuilder` re-animates from the current value toward the new end whenever the tween's `end` changes, which is one of its main selling points. Worth a half-sentence.

bdero (approved):
LGTM!

PR #602, opened 2026-07-28
chore: update Flutter to 3.44.8

nickshorebird:
## Summary

- Bump the documented Flutter version to 3.44.8, shipped in shorebird_cli 1.6.115.
- Dart stays at 3.12.2. The 3.44.7 -> 3.44.8 bump carries no Dart base change, so `dart_revision` is untouched.

## Testing

- Only `src/consts.ts` changed, single version string.


nickshorebird:
Version bump - Force merging

PR #607, opened 2026-08-06
docs: add Flutter add-to-app concepts guide

AbhishekDoshi26:
## Summary

Adds a new Flutter Concepts page covering the add-to-app pattern: embedding Flutter as a module inside an existing native Android or iOS app.

### New article: `flutter-concepts/add-to-app.mdx`

Covers:

- **Why add-to-app**: hybrid navigation stacks and partial-screen view use cases
- **Flutter module vs Flutter app**: structural differences, the `.android/`/`.ios/` generated wrappers, and the `@pragma('vm:entry-point')` requirement
- **FlutterEngine lifecycle**: pre-warming in `Application.onCreate()` / `AppDelegate`, with Kotlin and Swift examples
- **Multiple engines with `FlutterEngineGroup`**: explains the 100x memory difference between plain `FlutterEngine()` (~19 MB on Android, ~13 MB on iOS per engine) and `FlutterEngineGroup` (~180 KB per additional engine), with code samples for both platforms
- **Android integration**: source vs AAR Gradle dependency options; `FlutterActivity`, `FlutterFragment`, and `FlutterView` display mechanisms
- **iOS integration**: Swift Package Manager (recommended since Flutter 3.44+) vs CocoaPods; full-screen and partial-screen `FlutterViewController` patterns
- **Platform channels**: `MethodChannel` with a sequence diagram and Dart/Kotlin/Swift code examples
- **Multi-engine vs multi-view**: comparison table covering platforms, state sharing, and memory cost
- **Performance considerations**: warm-up timing, Impeller for first-frame jank, and a per-engine memory cost table

### Cross-linked from existing docs

- `code-push/index.mdx`: "Add-to-App (Hybrid Apps)" row in the feature support matrix now links to the new article
- `code-push/guides/hybrid-apps/android.mdx`: intro paragraph links to the new article for readers new to the pattern
- `code-push/guides/hybrid-apps/ios.mdx`: same


src/content/docs/code-push/guides/add-to-app/android.mdx
```diff
@@ -10,7 +10,9 @@ import { Image } from 'astro:assets';
 import hybridAndroidScreenshot from '~/assets/hybrid_android_screenshot.png';
 
 This guide explains how to use Shorebird in an Android hybrid app scenario (that
-is, your app embeds Flutter UI in non-Flutter UI).
+is, your app embeds Flutter UI in non-Flutter UI). If you are new to this
```
tomarra, line 13: Should we consider changing this section to be called "Add to App" instead of "Hybrid Apps"? I believe that "Add to App" is the more well known phrase for this but I could be wrong.

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,502 @@
...
+  Flutter widgets side by side. The Flutter portion renders one component (a
+  chart, a form, a product card) while the rest of the screen stays native.
+
+Beyond UI, add-to-app also lets you run shared non-UI Dart logic (networking,
+business rules, cryptography) inside a native app, taking advantage of Dart's
+portability and interoperability with other languages.
+
+## Flutter module vs Flutter app
+
+When you create a standalone Flutter app (`flutter create my_app`), the project
```
tomarra:
```suggestion
When you create a standalone Flutter app (`shorebird create my_app`), the project
```

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,502 @@
...
+contains a complete `android/` and `ios/` host wrapper that Flutter owns
+entirely. For add-to-app, you create a **Flutter module** instead:
+
+```bash
+flutter create --template=module my_flutter_module
```
tomarra: Do we support this via our `create`? Honestly not sure. If so should we use `shorebird` instead of `flutter` here?

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,502 @@
...
+
+Beyond UI, add-to-app also lets you run shared non-UI Dart logic (networking,
+business rules, cryptography) inside a native app, taking advantage of Dart's
+portability and interoperability with other languages.
+
+## Flutter module vs Flutter app
+
+When you create a standalone Flutter app (`flutter create my_app`), the project
+contains a complete `android/` and `ios/` host wrapper that Flutter owns
+entirely. For add-to-app, you create a **Flutter module** instead:
```
tomarra: Where should this command be run from? Inside the main project? Some other layer?

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,502 @@
...
+    main.dart          # Dart entry point
+  pubspec.yaml
+  .android/            # Auto-generated Android host (for standalone testing)
+  .ios/                # Auto-generated iOS host (for standalone testing)
+```
+
+The `.android/` and `.ios/` directories are generated wrappers that let you run
+the module in isolation with `flutter run`. They are prefixed with a dot to
+signal that they are managed by Flutter, so never edit them directly. When the
+module is embedded into your real native app, these wrappers are ignored. The
```
tomarra: Should we also say "not compiled into your app binary" here just so people don't think it's adding extra bloat?

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,502 @@
...
+        flutterEngine.run()
+        GeneratedPluginRegistrant.register(with: self.flutterEngine)
+        return super.application(application, didFinishLaunchingWithOptions: launchOptions)
+    }
+}
+```
+
+### Multiple engines
+
+When you need to run more than one Flutter screen simultaneously, use
```
tomarra: What would be the use case here? Assuming multiple windows or multiple flutter sections in a native view? It may be good to provide some context for this because you may not know that this is what you need until you run into it.

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,502 @@
...
+
+Each engine produced by `FlutterEngineGroup` is still an independent Dart
+program with its own isolate, state, and plugin registrations. The shared
+resources are managed internally and do not affect isolation between engines.
+
+## Android integration
+
+### 1. Add the module as a Gradle dependency
+
+There are two ways to depend on the Flutter module from your Android project.
```
tomarra: Should we be giving a preference to option A over B since I dont think our solution currently works with B given some customer feedback.

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,502 @@
...
+directory to generate an XCFramework:
+
+```bash
+flutter build ios-framework --xcframework --output=../MyFlutterPackage
+```
+
+Then add the local package in Xcode: **File > Add Package Dependencies > Add
+Local**, and point to the generated `MyFlutterPackage` directory.
+
+**CocoaPods setup** (alternative):
```
tomarra: Think we should remove this if it's truly in maintenance mode. The folks who have this setup already dont need it and if people are reading it for the first time just direct them to the new/correct way to do it.

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,502 @@
...
+addChild(flutterVC)
+flutterVC.view.frame = CGRect(x: 0, y: 200, width: view.bounds.width, height: 300)
+view.addSubview(flutterVC.view)
+flutterVC.didMove(toParent: self)
+```
+
+## Platform channels
+
+Platform channels are how Dart and native code send messages to each other. The
+most common type is `MethodChannel`, which uses an asynchronous request/response
```
tomarra, line 338: Should probably link to some Dart/Flutter docs for this as well.

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,502 @@
...
+    participant Dart
+    participant MethodChannel
+    participant Native
+
+    Dart->>MethodChannel: invokeMethod('getBatteryLevel')
+    MethodChannel->>Native: onMethodCall('getBatteryLevel')
+    Native-->>MethodChannel: result(42)
+    MethodChannel-->>Dart: Future resolves to 42
+```
+
```
tomarra, line 353: May be good to set the context for the use case you have laid out here. Getting battery level is a great example since we have to go to native code for that so explain it.

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,502 @@
...
+Start the `FlutterEngine` in `Application.onCreate()` (Android) or
+`application(_:didFinishLaunchingWithOptions:)` (iOS), not when the user
+triggers a navigation action. The first call to
+`dartExecutor.executeDartEntrypoint` is where the Dart VM initializes and the
+most latency is incurred.
+
+### First-frame jank
+
+Even with a pre-warmed engine, the first Flutter frame must compile all
```
tomarra: Since it's on by default do we need this? New users are going to get this for free.

tomarra (commented):
Some nits but mostly ready to go.

src/content/docs/code-push/guides/add-to-app/android.mdx
AbhishekDoshi26: Definitely! Add to app is more common term compared to hybrid apps

src/content/docs/flutter-concepts/add-to-app.mdx
AbhishekDoshi26: Yes we do. I just checked

src/content/docs/flutter-concepts/add-to-app.mdx
AbhishekDoshi26: Added use case

src/content/docs/flutter-concepts/add-to-app.mdx
AbhishekDoshi26: Oh right. Then I should remove option B

src/content/docs/flutter-concepts/add-to-app.mdx
AbhishekDoshi26: Added

src/content/docs/flutter-concepts/add-to-app.mdx
AbhishekDoshi26: Added

src/content/docs/flutter-concepts/add-to-app.mdx
AbhishekDoshi26: Removed

src/content/docs/code-push/guides/add-to-app/_meta.yml
```diff
@@ -0,0 +1,2 @@
+label: Add to App
```
tomarra: It would probably make sense to add redirects into [the config](https://github.com/shorebirdtech/docs/blob/main/astro.config.mjs) so the old paths resolve to the new one as this will change the url

src/content/docs/code-push/index.mdx
```diff
@@ -70,13 +70,13 @@ A typical Code Push workflow looks like this:
...
-| **Dart Code Patching (OTA)**         |   ✅    | ✅  |  ✅   |   ✅    |  ✅   |
-| **Flavors / Custom Targets**         |   ✅    | ✅  |  ✅   |   ✅    |  ✅   |
-| **Add-to-App (Hybrid Apps)**         |   ✅    | ✅  |  ❌   |   ❌    |  ❌   |
-| **Code Obfuscation (`--obfuscate`)** |   ✅    | ✅  |  ✅   |   ✅    |  ✅   |
-| **Patch Signing / KMS**              |   ✅    | ✅  |  ✅   |   ✅    |  ✅   |
+| Feature                                         | Android | iOS | macOS | Windows | Linux |
+| :---------------------------------------------- | :-----: | :-: | :---: | :-----: | :---: |
+| **Dart Code Patching (OTA)**                    |   ✅    | ✅  |  ✅   |   ✅    |  ✅   |
+| **Flavors / Custom Targets**                    |   ✅    | ✅  |  ✅   |   ✅    |  ✅   |
+| [**Add to App**](/flutter-concepts/add-to-app/) |   ✅    | ✅  |  ❌   |   ❌    |  ❌   |
```
tomarra: Style wise should this be 
```suggestion
| [**Add-to-app**](/flutter-concepts/add-to-app/) |   ✅    | ✅  |  ❌   |   ❌    |  ❌   |
```

tomarra: Confirm with @dawn-ducky and if so ensure we have this across the change

tomarra: Basing this off of https://docs.flutter.dev/add-to-app

src/content/docs/code-push/guides/add-to-app/_meta.yml
AbhishekDoshi26: Thank you! Added :)

src/content/docs/flutter-concepts/add-to-app.mdx
```diff
@@ -0,0 +1,486 @@
...
+
+### Texture and platform view overhead
+
+When Flutter renders inside a native view hierarchy, it composites its texture
+with the native rendering layer. On Android, this uses a `SurfaceTexture` or
+`SurfaceView` depending on the display mode. On iOS, it uses a `CALayer`. The
+compositing step adds a small GPU cost; for partial-screen views where Flutter
+covers only a fraction of the screen, this cost is negligible.
+
+## Further reading
```
dawn-ducky, line 475: All of these were referenced in the body of the article. If there are relevant Shorebird docs those can be added here if not this section should be removed.

dawn-ducky (changes requested):

src/content/docs/flutter-concepts/add-to-app.mdx
AbhishekDoshi26: removed

PR #631, opened 2026-08-25
docs: add "Adding Members" Arcade walkthrough to Organizations page

AbhishekDoshi26:
## What
- Imported the `<ArcadeEmbed />` component into `src/content/docs/account/orgs.mdx`.
- Added a new Arcade walkthrough video demonstrating how to add members to an organization.

## Why
Adding visual guides directly into the documentation makes it significantly easier for users to follow along with organization management features in the Shorebird Console. Consistent with our internal documentation rules, the video is placed prominently at the top of the `### Add a member` section so users get the visual context first.

Arcade: https://app.arcade.software/share/Cl6LTDv4A3TKhJJOdJse


dawn-ducky (changes requested):
The arcade is good, but it looks like the console has changed from what is described in the docs. The docs and screenshots  says "Add Member" if we're adding an arcade we should also update the images and copy to reflect the current state of the console.

AbhishekDoshi26:
> The arcade is good, but it looks like the console has changed from what is described in the docs. The docs and screenshots says "Add Member" if we're adding an arcade we should also update the images and copy to reflect the current state of the console.

Updated screenshots for the entire page :)

PR #632, opened 2026-08-25
docs: make Flutter Versions a general Flutter reference

eseidel:
## Status

**READY**

## Description

`/getting-started/flutter-version/` ranks around position 8 for the "what Flutter version is current" query cluster, but the page answers a Shorebird-specific question instead. This restructures it so the top half is useful to any Flutter developer and the Shorebird material sits below it.

### Page

- **Leads with the current stable release** in a callout, as one plain sentence rather than a stat block, so a search engine or assistant can lift it directly.
- **Tightened the recommendation prose** to three short paragraphs: latest stable, run the newest patch of whatever you're on, and the one-release-behind case that large enterprises use.
- **New release table** — newest patch of each recent release, its Dart SDK version, when it first shipped, and the headline changes it brought. Five rows show by default with a "Show 5 older releases" toggle. All ten rows ship in the HTML and collapse client-side, so crawlers and assistants read the whole table and it degrades to fully expanded with JS off.
- **Existing Shorebird version notes are unchanged**, now under a new "Shorebird and Flutter versions" heading. `#supported-flutter-versions` and `#use-a-different-flutter-version` keep their ids — Google indexes both as separate results.

### Data

- `scripts/update-flutter-releases.mjs` merges Flutter's three platform release feeds into `src/data/flutter-releases.json` (one row per release, newest patch, Dart version, dates). Run it with `npm run update-flutter-releases`. No generation timestamp, so a no-op run produces no diff.
- `.github/workflows/flutter-releases.yaml` runs daily and **fails** when the committed file has fallen behind stable, with a message naming the new version. It does not open a PR. Both paths were exercised locally.
- Highlights are hand-written in `src/data/release-highlights.ts`, summarized from Flutter's own release announcements — the feeds carry versions and dates but nothing about what changed. A release with no entry renders a dash.

Note that this is all separate from `src/consts.ts`, which tracks the Flutter version *Shorebird* ships and is still bumped by hand. The page distinguishes the two.

### Checks

`npm run build`, `format:check`, cspell, Vale, and the component-label lint all pass locally. Verified the table and callout in light and dark, and the toggle in a browser.

### Follow-ups, not in this PR

- Per-release "known issues" commentary sourced from the version-quality data in the admin dashboard, rather than the static notes list.
- Disclosure triangles for per-release detail — the data is already keyed per release.

### DIFF
--- a/.cspell.yaml
+++ b/.cspell.yaml
@@ -40,12 +40,14 @@ words:
   - cloudkms
   - codesign
   - codesigning
+  - cupertino
   - elif
   - evenodd
   - ffigen
   - fintech
   - frontmatter
   - gallego
+  - gles
   - graphql
   - geocodes
   - hotfixes
@@ -94,8 +96,11 @@ words:
   - shellenv
   - shorebirdtech
   - softwareupdate
+  - squircle
+  - squircles
   - subclassing
   - subosito
+  - superellipse
   - subview
   - tabler
   - temurin
--- a/.github/workflows/flutter-releases.yaml
+++ b/.github/workflows/flutter-releases.yaml
@@ -1,0 +1,52 @@
+name: flutter-releases
+
+# Fails when src/data/flutter-releases.json has fallen behind Flutter's release
+# feeds - a new stable, or a new patch on a line the table lists. The fix is to
+# run `npm run update-flutter-releases` and commit the result.
+
+on:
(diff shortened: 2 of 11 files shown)


AbhishekDoshi26 (approved):
LGTM 👍🏻

PR #636, opened 2026-08-29
docs(saml): either signing level, email fallbacks, account carry-over, rotation

eseidel:
The SAML guide told customers to configure around limitations the auth service no longer has, and stated a few of them as requirements. This removes those and leaves only the places where Shorebird is non-standard. The page says what to do, not what changed.

**Merge after** the two still-open auth PRs whose behaviour it now assumes — existing-account carry-over (shorebirdtech/_shorebird#2840) and metadata re-import (shorebirdtech/_shorebird#2847) — have landed.

## Removed, because standard SAML now just works

- "Shorebird requires both the response and the assertion to be signed" and the two `Signed (RSA-SHA256)` rows. Either level is accepted; every provider's default works.
- "The email attribute is required … sign-in fails without one of them, even when the Name ID is the email address." Okta's and Entra's defaults are read as-is.
- "Mention any members who already have a Shorebird account … those accounts need to be migrated … their first SSO sign-in fails." First SSO sign-in carries the account over.

## Kept or added, because Shorebird is non-standard there

- **Assertion encryption must be off** — new row. No decryption key is configured, so an IdP with encryption on fails; this was true before and undocumented.
- **Signed authentication requests must not be required** — unchanged.
- **Identity is keyed by email** — retitled section. A `Persistent`/`Transient` Name ID is opaque and needs an `email` attribute alongside it; the attribute table stays for that case and for the Okta walkthrough.
- **Metadata is not re-read automatically** — the rotation section now leads with that, since it is the one thing a customer has to act on: a certificate that was listed at the last read needs nothing, one generated since needs an email.

Troubleshooting's "SSO login failed" row no longer leads with a missing `email` attribute.

Prettier-formatted (`proseWrap: always`).

https://claude.ai/code/session_01Ks1X8BmfFzfJRm7nXJa7Vc


AbhishekDoshi26 (approved):

eseidel:
Merged `main` and resolved the conflict; also carried in one unrelated line, explained below.

**Conflict resolution.** #635 and #637 landed the metadata-URL pass on `main` and touched the two places this branch also touches:

- The account carry-over paragraph *moved* into "What to send Shorebird" on `main`, so the merge reintroduced a paragraph this PR deletes. Deleted again.
- The rotation paragraph was rewritten on both sides. Kept this branch's lead ("Shorebird does not re-read your metadata on its own") and folded in `main`'s hand-sent-values case, which this branch's wording had compressed away: *"If your connection was created from values sent by hand, send the new certificate as PEM text instead, and the old one too if it is still in use."*

**The `bingbot` line in `.cspell.yaml` is not part of this change.** `public/robots.txt` (#639) added a `Bingbot` user-agent line without the dictionary entry, so `spell-check / build` fails on `main` at 801470e and on every branch cut from it — this one included. This is the only open PR, so there was no fix to wait for. Since this PR already edits `.cspell.yaml`, the one-word addition rides along here and clears `main` when it merges. Happy to split it out if you'd rather it land on its own.

Local verification: `npm run build` (astro check, build, link validation) passes, `prettier --check` clean, `cspell` clean across all 102 files. `vale` I couldn't run — its binary doesn't install in this sandbox — so `style-check` in CI is the first real run of it.

One thing I couldn't check: the PR body says to merge after shorebirdtech/_shorebird#2840 and #2847 land. That repo isn't in this session's scope, so that gate is still yours to confirm.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #640, opened 2026-09-01
Feat: Documentation for Code Push API

dawn-ducky:
To make our docs more AI/agent friendly added documentation for Shorebird Code Push API, including authentication methods, resource model, endpoint groups, and common patterns.

Includes a direct link to the specification and built with inspiration from: https://vercel.com/docs/rest-api

## Status

**READY/IN DEVELOPMENT/HOLD**

## Description

Built a new page with a direct link to our API spec. The page includes some basic information about the spec, the spec should be the source of truth.


src/content/docs/account/api.mdx
```diff
@@ -0,0 +1,85 @@
+---
+title: API Keys
```
nickshorebird: I believe this will collide w/ the `API Keys` [api_keys.mdx](https://github.com/shorebirdtech/docs/blob/main/src/content/docs/account/api-keys.mdx) title already in account. No idea what will be served or if pages will just refuse. Suggest changing to `API Reference` or the like.

src/content/docs/account/api.mdx
```diff
@@ -0,0 +1,85 @@
...
+```
+Authorization: Bearer <token>
+```
+
+There are two kinds of tokens, and which one you use depends on context:
+
+- **API keys** (`sb_api_*`): long-lived,
+  [generated from the Shorebird console](https://docs.shorebird.dev/account/api-keys/).
+  Use these for CI pipelines, scripts, and any non-interactive automation. They
+  don't expire and aren't refreshed automatically, so treat them like any other
```
nickshorebird: This is a discouraged option per our UI. Default is 1 year w/ other options being 30 and 90 days.

nickshorebird (changes requested):
LGTM otherwise. Thanks for putting this together.

src/content/docs/account/api.mdx
dawn-ducky: Yup, this is what happens when I copy from another page and forget to change all the fields.

src/content/docs/account/api.mdx
dawn-ducky: @nickshorebird is this accurate:
"Use these for CI pipelines, scripts, and any non-interactive automation. Default expiration is 1 year, with options for 30 days and 90 days. Treat these like any other long-lived secret."

nickshorebird: It is perfect. Thank you.

nickshorebird (approved):
LGTM - Thanks for the changes.

AbhishekDoshi26 (approved):

PR #653, opened 2026-09-08
docs: guide for single-arch Android APKs (shorebirdtech/shorebird#3402)

### LINKED ISSUE #470: build: Build Apk and exclude somes Abi
I need to build my Apk only with one Abi (arm64).

In the default Flutter build, it is possible, but i cant find a way to do it with shorebird. I need to to reduce the size of my apk.

AzazelSensei:
I added a short guide for building a single-arch Android APK with Shorebird.

`shorebird release android --artifact apk --target-platform android-arm64` has to go before `--`. Gradle NDK ABI filters alone still make Shorebird look for every default architecture.

Also documented `shorebird releases get-apks --no-universal` so you get per-ABI APKs instead of only `universal.apk`.

shorebirdtech/shorebird#3402
Fixes #470

### DIFF
--- a/src/content/docs/code-push/faq.mdx
+++ b/src/content/docs/code-push/faq.mdx
@@ -461,8 +461,9 @@ can reproduce a rough measurement on your own app by comparing the output of
 
 ### How do you build a smaller APK for a single Android architecture?
 
-Use `--target-platform` on `shorebird release android`. Gradle NDK ABI filters
-alone do not restrict the architectures Shorebird uploads. See
+Use `--target-platform` on `shorebird release android` to declare which
+architectures to build. A Gradle NDK ABI filter that drops an ABI still listed
+in `--target-platform` is skipped, not treated as an error. See
 [Build a smaller APK for one architecture](/code-push/release/#build-a-smaller-apk-for-one-architecture).
 
 ### When do updates happen?
--- a/src/content/docs/code-push/faq.mdx
+++ b/src/content/docs/code-push/faq.mdx
@@ -459,6 +459,12 @@ tracked in
 can reproduce a rough measurement on your own app by comparing the output of
 `flutter build apk --release` against `shorebird build apk --release`.
 
+### How do you build a smaller APK for a single Android architecture?
+
+Use `--target-platform` on `shorebird release android`. Gradle NDK ABI filters
+alone do not restrict the architectures Shorebird uploads. See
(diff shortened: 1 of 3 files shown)


AbhishekDoshi26:
Thanks for writing this up — the flag details (`--target-platform` values/default, `--no-universal` default, `--split-per-abi` being unsupported, the `--` placement) all check out against the current CLI source.

One claim looks stale, though. In `release.mdx`:

> Gradle NDK ABI filters (for example `ndk.abiFilters`) are not enough on their own. Shorebird still looks for every architecture in `--target-platform` (all three by default). If Gradle drops an ABI Shorebird was asked to upload, the release fails. Keep those filters in sync with `--target-platform`, or drop the Gradle filters and use the flag only.

This describes the *pre-fix* behavior. `createAndroidReleaseArtifacts` in [`code_push_client_wrapper.dart`](https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart) now explicitly skips a missing arch directory instead of failing:

```dart
// Track which arch paths AGP didn't produce, so we can surface them at
// the end if zero archs uploaded. AGP omits an arch directory whenever
// the project filters it out via `ndk.abiFilters`, `splits.abi`, or
// `jniLibs.excludes`. Iterating only over present files lets a filtered
// release succeed instead of crashing on the first missing arch
// (https://github.com/shorebirdtech/shorebird/issues/3388).
```

It only errors if **zero** architectures are found for upload. Same pattern in `android_patcher.dart`'s `createPatchArtifacts`, referencing the same issue. That fix (#3388) landed and was closed 2026-05-03, before this PR.

So today: a Gradle-filtered ABI that's still in `--target-platform` doesn't fail the release — it's silently skipped and the release ships with only the ABIs that actually built. Might be worth documenting *that* instead (i.e. "use `--target-platform` to declare intent explicitly, since a mismatched Gradle filter won't be caught with an error").

AbhishekDoshi26 (changes requested):
Small change, rest looks good. Thanks!

AzazelSensei:
You're right — that was the old behavior, before #3388. I updated the Gradle filter note in `release.mdx` (and the matching FAQ line): a filtered ABI that's still in `--target-platform` is skipped, and the release ships with whatever actually built. `--target-platform` is how you declare intent; a mismatch no longer fails the release.

AbhishekDoshi26 (approved):
LGTM — fixed a Prettier formatting issue on release.mdx.

PR #654, opened 2026-09-09
feat: serve raw Markdown for every docs page at <path>.md

AbhishekDoshi26:
## Summary
- Adds `src/pages/[...slug].md.ts`, a catch-all endpoint that serves every docs page's source content as `text/markdown` at `<path>.md` (e.g. `/ci/view-logs.md`), so AI agents can fetch page content directly instead of scraping rendered HTML
- Reconstructs minimal frontmatter (title/description) and strips MDX component imports so the output reads as clean Markdown rather than raw source
- Complements the `public/_headers` rule (already on `main`) that sets `Content-Type: text/markdown` + CORS for `*.md` requests — that rule had no matching content until now, since neither `starlight-llms-txt`'s `/llms.txt`/`/llms-full.txt` routes nor Starlight itself serve per-page Markdown

## Test plan
- [x] `npm run build` succeeds (includes `astro check`)
- [x] Confirmed 70 `.md` files generated in `dist/`, one per non-draft docs page
- [x] Spot-checked `dist/index.md` and `dist/ci/view-logs.md` — clean frontmatter, imports stripped, Starlight aside syntax preserved
- [x] `npx prettier --check` passes on the new file
- [ ] After merge: verify `curl -I https://docs.shorebird.dev/ci/view-logs.md` returns `content-type: text/markdown` + `access-control-allow-origin: *` once deployed via CI

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,13 +10,19 @@
       "dependencies": {
         "@astrojs/check": "^0.9.10",
         "@astrojs/markdown-remark": "^7.3.0",
+        "@astrojs/mdx": "^7.0.5",
         "@astrojs/starlight": "^0.42.0",
         "@lavamoat/preinstall-always-fail": "^3.0.0",
         "@tailwindcss/vite": "^4.3.3",
         "astro": "^7.3.1",
         "astro-mermaid": "^2.1.0",
         "astro-opengraph-images": "^1.20.1",
+        "hast-util-select": "^6.0.4",
         "mermaid": "^11.17.2",
+        "rehype-parse": "^9.0.1",
+        "rehype-remark": "^10.0.1",
+        "remark-gfm": "^4.0.1",
+        "remark-stringify": "^11.0.0",
         "sharp": "^0.35.3",
         "starlight-auto-sidebar": "^0.4.1",
         "starlight-image-zoom": "^0.16.0",
@@ -24,10 +30,12 @@
         "starlight-theme-nova": "^0.12.3",
         "tailwindcss": "^4.2.0",
         "tw-to-css": "^0.0.12",
-        "typescript": "^6.0.3"
+        "typescript": "^6.0.3",
+        "unified": "^11.0.5"
       },
       "devDependencies": {
         "@lavamoat/allow-scripts": "^5.1.0",
+        "@types/hast": "^3.0.5",
         "@types/react": "^19.2.18",
         "@vvago/vale": "^3.17.1",
         "cspell": "^10.2.1",
--- a/package.json
+++ b/package.json
@@ -17,13 +17,19 @@
   "dependencies": {
     "@astrojs/check": "^0.9.10",
     "@astrojs/markdown-remark": "^7.3.0",
+    "@astrojs/mdx": "^7.0.5",
     "@astrojs/starlight": "^0.42.0",
     "@lavamoat/preinstall-always-fail": "^3.0.0",
     "@tailwindcss/vite": "^4.3.3",
     "astro": "^7.3.1",
     "astro-mermaid": "^2.1.0",
     "astro-opengraph-images": "^1.20.1",
+    "hast-util-select": "^6.0.4",
     "mermaid": "^11.17.2",
+    "rehype-parse": "^9.0.1",
+    "rehype-remark": "^10.0.1",
+    "remark-gfm": "^4.0.1",
+    "remark-stringify": "^11.0.0",
     "sharp": "^0.35.3",
     "starlight-auto-sidebar": "^0.4.1",
     "starlight-image-zoom": "^0.16.0",
@@ -31,10 +37,12 @@
     "starlight-theme-nova": "^0.12.3",
     "tailwindcss": "^4.2.0",
     "tw-to-css": "^0.0.12",
-    "typescript": "^6.0.3"
+    "typescript": "^6.0.3",
+    "unified": "^11.0.5"
   },
   "devDependencies": {
     "@lavamoat/allow-scripts": "^5.1.0",
+    "@types/hast": "^3.0.5",
     "@types/react": "^19.2.18",
     "@vvago/vale": "^3.17.1",
     "cspell": "^10.2.1",
--- a/src/pages/[...slug].md.ts
+++ b/src/pages/[...slug].md.ts
@@ -1,0 +1,35 @@
+import type { APIRoute, GetStaticPaths } from 'astro';
+import { getCollection } from 'astro:content';
+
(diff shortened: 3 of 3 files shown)


eseidel:
Hi Abhishek — this is Claude, reviewing on Eric's behalf.

Nice idea, and the `_headers` gap you spotted is real. I checked the branch out locally and ran `npm run build`, then diffed the generated `dist/**/*.md` against the source. Three issues, all from the same regex — `body.replace(/^import .+\n+/gm, '')` runs over the whole document with the `m` flag, so it matches any line starting with `import`, anywhere.

**1. It deletes Dart imports out of code samples.** This is a Flutter docs site, so 6 files have `import 'package:...'` at the start of a line inside a fence. From the build:

`src/content/docs/code-push/guides/staging-patches.mdx:92` has

    ```diff
    import 'package:flutter/material.dart';

    void main() {

and `dist/code-push/guides/staging-patches.md:85` comes out as

    ```diff
    void main() {

An agent that copies that sample gets code that doesn't compile. Same in `tracks.mdx`, `code-push-guide.mdx`, `add-to-app.mdx`, and both `state-management` pages.

**2. Multi-line imports get half-eaten.** `.+` matches `import {`, `\n+` eats the newline, and the rest of the statement survives. `dist/getting-started.md` — the Quick Start page — currently begins:

    ---
    title: "Quick Start"
    ---

    CodeTabs,
      LinkButton,
      LinkCard,
    } from 'starlight-theme-nova/components';
    <ArcadeEmbed

**3. Version placeholders leak.** The route serves `entry.body`, so the `remarkReplaceVersions` plugin never runs. `dist/code-push/release.html` says `--flutter-version 3.47.2`; `dist/code-push/release.md` says `--flutter-version %flutter_version%`. 6 files affected, and it's exactly the content an agent is most likely to paste into a terminal.

Minor: the `Content-Type` header on the `Response` is dead code — with `prerender = true` the route is written to disk as a file and the header is discarded. `public/_headers` is what actually applies. Worth a comment so nobody later "fixes" `_headers` by pointing at this line.

**On the approach.** There's no first-party Starlight support for this (`curl https://starlight.astro.build/getting-started.md` → 404), so a custom route is reasonable — I wouldn't take a dependency on one of the young community plugins for 35 lines. But the recommended technique is already in our `node_modules`: `starlight-llms-txt/entryToSimpleMarkdown.ts` renders the entry through Astro's real pipeline and converts the HTML back to Markdown:

    const { Content } = await render(entry);
    const html = await astroContainer.renderToString(Content, context);
    const markdown = String(await htmlToMarkdownPipeline.process(html));

That gets remark plugins (version substitution), resolved components, and intact code fences for free — no regex. It's why `/llms-full.txt` on the site is already correct where these `.md` files aren't. The package only exports `.`, so it means copying ~20 lines of pipeline setup rather than importing it.

If you'd rather keep the raw-source approach for a smaller diff, the minimum fix is to anchor the strip to the leading import block only — loop `/^import\s[^;]*?;\s*/` off the front, which handles multi-line statements and can never reach a code fence — and run the body through the version replacements. The output would still carry bare `<Tabs>`/`<ArcadeEmbed>` JSX and will drift from the rendered page again the next time a remark plugin is added.

Either way, a build-time assertion that no `%...%` placeholder survives into `dist/**/*.md` would be worth adding.

AbhishekDoshi26:
Good catches, thanks. Switched to the rendering-pipeline approach — pushed in 9ae3181.

- Dart imports in code fences: no longer touched, confirmed \`import 'package:flutter/material.dart';\` survives intact in \`dist/code-push/guides/staging-patches.md\`.
- Multi-line MDX imports: gone from output entirely now (they're compile-time only and never reach the rendered HTML), confirmed clean in \`dist/getting-started.md\`.
- Version placeholders: substituted correctly since \`render(entry)\` goes through the same pipeline as the HTML build — confirmed \`--flutter-version 3.47.2\` in \`dist/code-push/release.md\`, and grepped all of \`dist/**/*.md\` for any stray \`%...%\` pattern, none found.
- Dropped the now-genuinely-dead \`Content-Type\` header comment concern by leaving the header in place with a comment explaining \`public/_headers\` is what actually applies in production.

Skipped the minify/custom-selector/tabs-list/expressive-code parts of \`entryToSimpleMarkdown.ts\` since none apply here (site has \`expressiveCode: false\`, and we want full content per page, not the minified llms-small variant) — kept it to the core render → rehype-remark → remark-gfm → remark-stringify pipeline.

Didn't add the build-time \`%...%\` placeholder assertion — feels like a separate follow-up rather than blocking this PR, happy to add if you'd rather have it in here.

AbhishekDoshi26:
Ran a self-review on this. Found two more things worth fixing, pushed in 67e77f3:

**Fixed:**
- \`<Tabs>\` renders as a \`<starlight-tabs>\` custom element the generic HTML→Markdown conversion doesn't understand — tab labels were getting dropped and every panel's content ran together with dead links to nonexistent anchors. \`/code-push/release.md\` mixed Android/iOS/Linux/macOS/Windows release instructions in one unlabeled stream — a real misattribution risk for exactly the agents this endpoint is for. Added the same \`<starlight-tabs>\`→list conversion \`starlight-llms-txt\` uses for \`/llms-full.txt\`. Verified: all 5 tabs on that page now render as labeled list items.
- \`@astrojs/mdx\`, \`hast-util-select\`, \`rehype-parse\`, \`rehype-remark\`, \`remark-gfm\`, \`remark-stringify\`, \`unified\` were only resolving because npm happened to hoist them as transitive deps of \`starlight-llms-txt\`/\`@astrojs/starlight\`, undeclared in \`package.json\`. Declared them explicitly (plus \`@types/hast\` as a devDependency) so a future lockfile regen can't silently break the build.

**Not fixed, flagged instead:** two more content-quality issues, both pre-existing in the already-shipped \`/llms-full.txt\` (same rendering technique) rather than something this PR introduced — not a regression, so didn't fix here:
- \`LinkCard\` links run straight into their description text with no separator (\`[Getting Started](/getting-started/)Install Shorebird...\`)
- Starlight's sr-only heading-permalink text leaks as a visible duplicate line under every heading (\`[Section titled "X"](#x)\`)

Both would be better fixed once in \`starlight-llms-txt\`'s shared pipeline than duplicated here — happy to open that as a separate issue/PR if useful.

Also verified: Mermaid diagrams lose their \`mermaid\` language tag on conversion (also pre-existing in \`/llms-full.txt\`), and the module-level \`astroContainer\` singleton mirrors the exact pattern \`entryToSimpleMarkdown.ts\` already uses — not treating either as blocking.

PR #655, opened 2026-09-09
docs: name the API catalog and say what the API guarantees

eseidel:
Companion to shorebirdtech/_shorebird#3020, which makes the API reachable by callers that are not `shorebird_code_push_client` and adds an RFC 9727 catalog at `/.well-known/api-catalog`.

Three changes to the API reference page:

- Points at the catalog, for a reader who has the hostname but not this page.
- States that the API carries no compatibility guarantee yet, matching the wording now in the spec's own `info.description`. Whether we commit to a stable public API is a separate decision; this says where we actually stand today.
- Notes that a bearer token is the only header a request needs. `x-version` is `shorebird_code_push_client`'s own `packageVersion` and was never something a third party could send meaningfully; after #3020 they no longer have to.

Should be merged after shorebirdtech/_shorebird#3020
https://claude.ai/code/session_01LG6NjeENXsouDyHrt1YPB7


AbhishekDoshi26:
Thanks for adding this — a couple of things worth checking before merging:

**"No other headers are required" doesn't match what the API actually does.** I tested this live against `api.shorebird.dev`: every `/api/v1/*` endpoint (aside from the exempted `patches/check`) also requires an `x-version` header. Without it, you get a 426 with `"Missing x-version header"` even with a valid bearer token:

```
$ curl -H "Authorization: Bearer <token>" https://api.shorebird.dev/api/v1/apps
{"code":"upgrade_required","message":"The current version is not supported...","details":"Missing x-version header."}
```

**The new API-catalog link hits the same wall.** Loading `https://api.shorebird.dev/.well-known/api-catalog` plainly — exactly the scenario this section describes (someone with just the hostname) — returns the same 426/`Missing x-version header` error instead of the catalog.

Might be worth confirming with the API team whether `x-version` is meant to be documented here, or whether the catalog endpoint should be exempt from that check (like `patches/check` is for auth). As written, both new/changed sentences will send readers into an error they're not warned about.

Minor: "You are welcome to build on it; expect to follow it" reads a bit ambiguously — worth rephrasing to state plainly that the API can change without notice.

AbhishekDoshi26 (approved):
LGTM! Should be merged after shorebirdtech/_shorebird#3020

PR #656, opened 2026-09-10
feat: content-negotiate Markdown for AI agents on canonical URLs

AbhishekDoshi26:
## Summary
- Adds `functions/_middleware.ts`, a Cloudflare Pages Function that serves a page's Markdown sibling when the request's `Accept` header prefers `text/markdown` over `text/html` — on the same canonical URL a person would visit (e.g. `GET /ci/view-logs/` with `Accept: text/markdown` returns Markdown), with `Vary: Accept` on every negotiated response.
- The `<path>.md` URLs from #654 are unaffected; this adds negotiation on the canonical URL on top, which is what the Vercel and Cloudflare agent-friendliness scans check for.
- Falls back to HTML when a route has no Markdown sibling (e.g. non-docs pages), and only returns `406` when a client explicitly rejects every representation (`Accept: */*;q=0`).

## Why not Cloudflare's "Markdown for Agents" toggle
That's a dashboard setting under AI Crawl Control, but it's zone-level and requires the domain to be on Cloudflare DNS with a Pro/Business plan. Checked both Cloudflare accounts we have access to — neither has `shorebird.dev` as an actual zone, just the Pages custom domain via an external CNAME — so that toggle isn't reachable without a larger DNS migration. This reimplements the negotiation logic ourselves instead, following [acceptmarkdown.com's Cloudflare Workers recipe](https://acceptmarkdown.com/recipes/cloudflare-workers), adapted to a Pages Function rather than the from-scratch Workers-with-Assets setup the recipe describes, since we're already deploying via `wrangler pages deploy`.

## Notes
- `functions/` is excluded from the root `tsconfig.json` (so `astro check` skips it) since `@cloudflare/workers-types`' `Request`/`Response`/`Headers` overrides collide with the DOM lib types the rest of the project uses. It gets its own scoped `functions/tsconfig.json` instead, for editor support.

## Test plan
- [x] `npx tsc --noEmit -p functions/tsconfig.json` passes
- [x] `npm run build` (includes `astro check`) passes
- [x] Verified locally with `wrangler pages dev dist`:
  - [x] Browser-style `Accept` → HTML + `Vary: Accept`
  - [x] `Accept: text/markdown` → Markdown on the canonical URL, `content-type: text/markdown; charset=utf-8`
  - [x] No `Accept` header → HTML (safe default)
  - [x] Root `/` with `Accept: text/markdown` → `/index.md` content
  - [x] Competing q-values both directions (`text/markdown;q=0.9` beats `text/html;q=0.5` and vice versa)
  - [x] `Accept: */*;q=0` → `406`
  - [x] A route with no Markdown sibling (`/roadmap/`) → graceful HTML fallback, no error
  - [x] Explicit `.md` URLs and static assets unaffected
- [ ] After merge: verify on `docs.shorebird.dev` directly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@ dist/
 # generated types
 .astro/
 
+# local wrangler dev state
+.wrangler/
+
 # dependencies
 node_modules/
 
--- a/functions/_middleware.ts
+++ b/functions/_middleware.ts
@@ -1,0 +1,131 @@
+/// <reference types="@cloudflare/workers-types" />
+
+// Content negotiation for AI agents: a request for a normal docs URL with
+// `Accept: text/markdown` gets that page's Markdown sibling instead of HTML,
+// on the same URL a person would visit. The `.md` URLs themselves (added in
+// #654) keep working unchanged — this adds the "ask the canonical URL for
+// Markdown" path on top, which is what agent-friendliness scanners check for.
+//
+// Cloudflare's dashboard-level "Markdown for Agents" (AI Crawl Control) does
+// the same thing, but only for a zone on Cloudflare DNS with a Pro/Business
+// plan — shorebird.dev isn't (DNS lives elsewhere, just CNAMed to Pages), so
+// this reimplements the negotiation ourselves. Recipe:
+// https://acceptmarkdown.com/recipes/cloudflare-workers
+
+interface Env {
+  ASSETS: Fetcher;
+}
+
+// Static assets and non-content routes never have a Markdown sibling; skip
+// negotiation for them entirely.
+const SKIP_PATTERN =
+  /\.(css|js|mjs|json|png|jpe?g|gif|svg|webp|avif|ico|woff2?|ttf|eot|pdf|xml|txt|map)$/i;
+
+function markdownSiblingPath(pathname: string): string {
+  const trimmed =
+    pathname.endsWith('/') && pathname !== '/'
+      ? pathname.slice(0, -1)
+      : pathname;
+  return trimmed === '/' ? '/index.md' : `${trimmed}.md`;
+}
+
+interface AcceptEntry {
+  type: string;
+  q: number;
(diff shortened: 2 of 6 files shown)


AbhishekDoshi26:
Reviewed this myself with fresh eyes. Found four real bugs, pushed in f521e8e:

- **Wildcard masked an explicit rejection**: \`Accept: text/html;q=0, */*\` still served HTML — \`scoreFor()\` took the max across all matching entries, so the \`*/*;q=1\` match overrode the exact \`text/html;q=0\` refusal. Per RFC 9110 §12.5.1, q=0 on the exact type is a hard exclusion a wildcard can't override. Now checked first.
- **Whitespace around \`=\` broke q parsing**: \`text/markdown; q = 0.5\` parsed as q=1 (the split key came out as \`"q "\` with a trailing space, failing the \`=== 'q'\` check), which could flip a negotiation that should've gone the other way. Now trims both sides.
- **\`Vary\` handling was inconsistent between branches**: the markdown branch used \`.set()\` (discards any \`Vary\` the edge/asset layer already set, e.g. \`Accept-Encoding\`), the HTML branch used \`.append()\` (risks a duplicate \`Accept\` token). Unified into one helper that merges correctly either way.
- **HEAD requests got a body**: the markdown branch always issued a GET to fetch the \`.md\` sibling and returned its full body regardless of the original method, so \`HEAD\` + \`Accept: text/markdown\` got content it didn't ask for. Now forwards the original method and nulls the body for HEAD.

Verified all four against the running negotiation logic via \`wrangler pages dev\` — confirmed each reproduced before the fix and is resolved after, then reran the original scenario suite to confirm no regressions. One caveat: the HEAD fix can't be fully observed locally — \`wrangler pages dev\`'s static asset layer returns a body on HEAD even for a direct, unmodified request to a raw \`.md\` file, confirmed independent of this code — but the response construction is correct now regardless of that local simulator gap.

PR #657, opened 2026-09-11
feat: improve docs readiness for AI agents

AbhishekDoshi26:
## Summary

Adds agent-readiness and discovery metadata to `docs.shorebird.dev`:

- **Link response headers (`functions/_middleware.ts`)**: Injects RFC 8288 `Link` headers on documentation responses advertising `api-catalog`, `agent-skills`, `agent.json`, `oauth-authorization-server`, `oauth-protected-resource`, `openid-configuration`, `llms.txt`, `opensearch.xml`, `humans.txt`, and per-page markdown alternatives.
- **Content negotiation & 406 handling (`functions/_middleware.ts`)**:
  - Serves Markdown twins with `Vary: Accept` on canonical URLs when requested with `Accept: text/markdown`.
  - Returns `406 Not Acceptable` with `Vary: Accept` when an unsupported MIME type is requested (e.g. `Accept: application/x-content-negotiation-probe`) per RFC 9110 and AcceptMarkdown readiness tests.
- **Agent-friendly 404 recovery**:
  - Adds `src/content/docs/404.md` linking to `llms.txt`, `llms-full.txt`, sitemaps, API spec, and homepage.
  - Updates `functions/_middleware.ts` so 404s return structured markdown when requested with `Accept: text/markdown`, or structured JSON with `resolution_hints` on `Accept: application/json`.
- **Discovery manifests (`public/.well-known/`)**:
  - `api-catalog` & `api-catalog.json`: RFC 9727 linkset pointing to API endpoints, OpenAPI 3.1 spec, and reachability status.
  - `agent-skills/index.json`: Manifest defining CLI workflows across Android, iOS, macOS, Windows, and Linux.
  - `agent.json`: Capability card for A2A discovery.
  - `oauth-authorization-server`, `oauth-protected-resource`, `openid-configuration`: OAuth 2.0 / OIDC metadata for API key and token discovery.
  - `security.txt`: RFC 9116 vulnerability reporting policy.
  - `traffic-advice`: Crawler guidance.
  - `agent-instructions.txt`: Guidance on CLI usage, releases vs. patches, and multi-platform support.
  - `/auth.md`: Authentication reference for API keys and tokens.
- **In-browser WebMCP tool provider (`src/components/starlight/Head.astro`)**: Exposes client tools (`get_install_command`, `get_llms_txt_url`) on `window.__modelContextProtocol` and dispatches `mcp:ready`.
- **OpenSearch & Humans.txt**: Adds `public/opensearch.xml` for agent search and `public/humans.txt` for team/standards attribution.
- **Schema.org JSON-LD graph**: Adds structured data in `Head.astro` linking to Flutter (`Q56276200`) and Dart (`Q248160`) Wikidata entities and founder Eric Seidel.
- **Crawl directives & CORS (`public/robots.txt`, `public/_headers`)**:
  - Explicitly allows major AI crawlers (`GPTBot`, `ClaudeBot`, `PerplexityBot`, `Google-Extended`, `Applebot-Extended`, `Cohere-ai`, `meta-externalagent`, `OAI-SearchBot`, `Bytespider`) with `Content-Signal: search=yes, ai-input=yes, ai-train=yes`.
  - Sets `Access-Control-Allow-Origin: *` and UTF-8 charset on `llms*.txt` and `sitemap*.xml`.

## DNS note (Wesley)

To enable DNS-AID discovery, add an HTTPS/SVCB record in Cloudflare DNS for `shorebird.dev`:
- Name: `_index._agents.docs`
- Type: `HTTPS`
- Target: `docs.shorebird.dev.`
- Priority: `1`
- Parameters: `alpn="h2,h3" port="443"`

## Test plan

- [x] `npm run build` succeeds (72 pages built cleanly, 0 broken links)
- [x] `npm run format:check` passes
- [x] `npx cspell --no-progress "**"` passes (0 issues)
- [x] `npx vale public src` passes (0 errors/warnings)
- [x] Verified locally: content negotiation (`Accept: text/markdown`, `406 Not Acceptable` on unsupported probe types, and `Accept: application/json` on 404s), RFC 8288 headers, and WebMCP tools
- [x] CI workflow checks passing on PR #657


AbhishekDoshi26:
Reviewed this with fresh eyes. Found two real regressions and three maintainability issues, pushed in 895f978. All verified against actual behavior via \`wrangler pages dev\`, not just read through.

**Regressions (both confirmed reproducing, both fixed):**

- **\`.well-known/*\` discovery files 406'd on their own declared Accept type.** These files have no extension, so \`SKIP_PATTERN\` never exempted them from negotiation — and \`negotiate()\` only understands \`text/markdown\`/\`text/html\`/\`application/json\`. So \`Accept: application/trafficadvice+json\` on \`/.well-known/traffic-advice\`, or \`application/linkset+json\` on \`api-catalog\`, scored negative everywhere and got a hard 406 — breaking the exact machine-verifiable check these files exist for. This would very plausibly re-fail the audit categories this PR targets, since testing a resource's own declared content type is exactly what a careful scanner does. Fixed by exempting \`/.well-known/*\` from negotiation entirely — they're single-format RFC 8615 discovery files, not negotiable docs content.
- **Narrow-but-not-hostile Accept headers now hard-406.** Adding JSON support dropped the "only 406 on an explicit \`*/*;q=0\` rejection" nuance — any Accept header that didn't happen to list one of our three types (e.g. \`Accept: text/plain\` from a health check) now 406'd instead of gracefully falling back to HTML. Restored the original semantics.

**Maintainability (all three had visible drift already):**

- \`api-catalog\` / \`api-catalog.json\` were byte-identical files that had to be hand-kept in sync — deduped into one canonical file + a \`_redirects\` 200-rewrite for the \`.json\` alias.
- \`MARKDOWN_404_BODY\` had already drifted from the real 404 page (missing the System Status and Agent Instructions links added since). Now fetches the actual \`/404.md\` the build generates instead of hand-maintaining a second copy.
- The markdown-sibling-path logic was duplicated between \`_middleware.ts\` and \`Head.astro\` — extracted to \`src/utils/markdown-path.ts\` so the \`<link rel="alternate">\` tag and the middleware's actual negotiation can't drift apart.

Also collapsed the repeated "copy headers, add Vary/Link, null body on HEAD" block (duplicated 6x across branches) into one \`respond()\` helper while I was in there.

No test framework exists in this repo to add automated coverage to — verified all of the above by reproducing each bug against the built \`dist/\` output before the fix and confirming it resolved after, then reran the full existing scenario suite (browser Accept, explicit markdown/json, competing q-values, HEAD requests, explicit full rejection, no-sibling fallback, static assets, existing \`.md\` URLs) to confirm no regressions.

eseidel:
Posting on Eric's behalf (Claude, from a session Eric is driving). Three of the `.well-known` files this PR added describe an authorization server that does not exist, and they should be removed rather than corrected.

**`oauth-authorization-server`, `openid-configuration`**

Every endpoint named here 404s on `auth.shorebird.dev`:

```
/oauth/authorize   404
/oauth/token       404
/oauth/userinfo    404
```

The auth service today has `/login` (human sign-in) and `/token` (authorization_code + PKCE, refresh_token). The other fields don't match the server either:

- `scopes_supported: apps:read, apps:write, releases:read, ...`: none of these scope names exist. Keys are minted as `full_access` or `release_and_patch`; login sessions carry `admin` or `preview_read`.
- `token_endpoint_auth_methods_supported: client_secret_post, client_secret_basic`: there are no client secrets. The server has no client registration yet.
- `registration_endpoint` points at a docs page. RFC 7591 registration is a POST endpoint on the issuer; a docs URL here will be called with a POST and fail.
- `response_types_supported: token`, `id_token`: there is no implicit flow and no OIDC. `openid-configuration` should not exist at all.
- `grant_types_supported: urn:ietf:params:oauth:grant-type:token-exchange`: not implemented.

RFC 8414 also fixes *where* this document lives: `{issuer}/.well-known/oauth-authorization-server`, i.e. `auth.shorebird.dev`, not `docs.shorebird.dev`. shorebirdtech/_shorebird#3031 adds the real document there, generated from the server's own route table and scope list. Once it deploys, agents will find two contradictory metadata documents.

**`oauth-protected-resource`**

Same problem: `scopes_supported` lists names the API does not recognise. RFC 9728 puts this at the resource server (`api.shorebird.dev`), and the real scopes are in `api.shorebird.dev/openapi.json` (shorebirdtech/_shorebird#3043).

**Fix**

Delete the three files, drop the three `rel=` entries from the `Link` header in `functions/_middleware.ts` (lines 82-84), remove the three entries from `api-catalog`, `api-catalog.json`, and llms.txt's Optional list, and remove the three URLs from `auth.md`'s "Protected resources" section.

**Minor, `auth.md`**: "Lifespan: Configurable for 30 days, 90 days, or 1 year." Keys can also be created with no expiry; that's the console default and the OpenAPI spec describes them as long-lived.

The rest of the PR (llms.txt, markdown negotiation, 404 handling, robots, agent-skills, agent.json) checked out against the live endpoints.

AbhishekDoshi26:
Good catch, and fair — my review of this PR checked the negotiation mechanism (headers, content-types, HEAD handling) but never verified the OAuth metadata's actual claims against the live \`auth.shorebird.dev\` server. Should have.

Rebased #658 onto #659: dropped the \`agent_auth\` block I'd added to \`oauth-authorization-server\` (nothing to add it to anymore), kept the two unrelated fixes (\`auth.md\` heading, \`ai-catalog.json\` identifier field). Verified the three deleted files correctly 404 now.

Also corrected our internal tracking doc, which had wrongly marked the OAuth/OIDC and Protected Resource Metadata items as shipped.

eseidel:
To be clear, I don't know any of this, claude does.  I just ask claude to do this stuff and then check its work.  I'm far from an oauth expert.  :)

PR #664, opened 2026-09-12
docs: document Android keystore setup in the GitHub CI guide

### LINKED ISSUE #338: Add instructions re: setting up Android keystore to github guide
https://docs.shorebird.dev/ci/github/ does not make any mention of Android keystores, which is a common pain point for our users.

eseidel:
## Status

**READY**

## Description

Fixes #338.

The GitHub Actions guide mentions signing once, in a YAML comment saying signing information must be set up on the runner, with no instructions for doing it.

Adds a short "Android code signing" section: `flutter build` does the signing, so it is standard Flutter — links to [Flutter's instructions](https://docs.flutter.dev/deployment/android#sign-the-app) and to ours — plus the one thing neither covers. On CI the keystore is not already on the machine, so store it as a base64 secret, decode it outside the checkout, and point `key.properties` at it. Then what happens if you skip it.

+17 lines, no YAML.

### What changed during review

Started at 90 lines with a full workflow, a secret-storage subsection, and two callouts. Now three paragraphs. Three things were wrong or unnecessary:

- The claim that `shorebird patch android` needs the same signing config. `AndroidArchiveDiffer.changedFiles` does report `META-INF/*` as changed, but warnings come from `containsPotentiallyBreakingAssetDiffs` (`assets`/`res`/`AssetManifest.json`) and `containsPotentiallyBreakingNativeDiffs` (`.dex`). `META-INF` matches neither.
- "Shorebird does not sign your app" as the framing. Gradle signs, during `flutter build`; Shorebird requires nothing.
- A hand-written recipe where links would do.

The earlier snippet also decoded the keystore into `android/app/`. `_zap`'s release workflow decodes to `$RUNNER_TEMP` instead, which is better, so the prose says "outside your checkout".

### On the standard-action question

There is no action for this half. Gradle signs during `flutter build`, so the keystore has to be in place first; the common actions (`r0adkll/sign-android-release` and similar) sign an artifact after it is built. `_zap` hand-rolls the decode step for the same reason and uses an action only for the Play upload.

### Verification

- `npx prettier --write` — clean
- `npx cspell --config .cspell.yaml` — 0 issues
- `vale 3.21.0 --config=.vale.ini src/content/docs` — 0 errors, matching the version CI installs
- `npx astro build` — all internal links valid

https://claude.ai/code/session_015fnEF9Ch8k5VQnwmBGD1nn

### DIFF
--- a/src/content/docs/code-push/ci/github.mdx
+++ b/src/content/docs/code-push/ci/github.mdx
@@ -6,7 +6,7 @@ sidebar:
   order: 8
 ---
 
-{/* cspell:words shorebirdtech */}
+{/* cspell:words shorebirdtech appbundle */}
 
 import Authentication from './_authentication.mdx';
 import { LinkCard } from 'starlight-theme-nova/components';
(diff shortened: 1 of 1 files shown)


src/content/docs/code-push/ci/github.mdx
```diff
@@ -87,6 +87,125 @@ secret: <THE GENERATED SHOREBIRD_TOKEN>
...
+Shorebird does not sign your app. `shorebird release android` and
+`shorebird patch android` run Gradle, and Gradle signs with whatever release
+signing configuration your project already has. A fresh GitHub runner has no
+keystore, so unless you put one there, Gradle falls back to the debug signing
+key and the resulting build is rejected by the Play Console or fails to install
+over an existing copy of your app with `INSTALL_PARSE_FAILED_NO_CERTIFICATES`.
+
+`shorebird patch android` needs this too. Patching rebuilds your app so it can
+be compared against the released version, so the runner needs the same signing
+configuration the release was built with.
```
eseidel: I dont' think this is true.  I'm pretty sure we strip the signature when comparing the patch.

src/content/docs/code-push/ci/github.mdx
```diff
@@ -87,6 +87,125 @@ secret: <THE GENERATED SHOREBIRD_TOKEN>
...
 authenticated functions such as creating patches 🎉
 
+## Android code signing
+
+Shorebird does not sign your app. `shorebird release android` and
+`shorebird patch android` run Gradle, and Gradle signs with whatever release
+signing configuration your project already has. A fresh GitHub runner has no
+keystore, so unless you put one there, Gradle falls back to the debug signing
+key and the resulting build is rejected by the Play Console or fails to install
+over an existing copy of your app with `INSTALL_PARSE_FAILED_NO_CERTIFICATES`.
```
eseidel: It's less that we "dont sign it" so much as we do the same work `flutter build` does, and signing is typically after that.

src/content/docs/code-push/ci/github.mdx
```diff
@@ -87,6 +87,125 @@ secret: <THE GENERATED SHOREBIRD_TOKEN>
...
+`ANDROID_KEY_PASSWORD`, and `ANDROID_KEY_ALIAS`.
+
+:::danger
+
+Never commit `upload-keystore.jks` or `android/key.properties` to your
+repository. Both belong in `.gitignore`. Losing the upload key means you can no
+longer update your app on the Play Store without contacting Google, so keep a
+backup somewhere safe.
+
+:::
```
eseidel: Why is any of this necessary?  Can't we link to some standard signing setup guide?  I'm not clear why we need one specific to us?  We set up zap with Android signing (private repo) can we just look at what we did there?

src/content/docs/code-push/ci/github.mdx
```diff
@@ -87,6 +87,125 @@ secret: <THE GENERATED SHOREBIRD_TOKEN>
...
+
+### Recreate the keystore on the runner
+
+Add a step that decodes the keystore and writes `android/key.properties` before
+any Shorebird step runs:
+
+```yaml
+- name: 🔐 Set up Android signing
```
eseidel: This seems wrong.  There must be a standard GHA to handle this.  This whole guide should be *much* shorter and probably just a link to some other standrad sign on android action and/or guide.

src/content/docs/code-push/ci/github.mdx
```diff
@@ -87,6 +87,118 @@ secret: <THE GENERATED SHOREBIRD_TOKEN>
...
 
+## Android code signing
+
+`shorebird release android` and `shorebird patch android` run Gradle, which
+signs with your project's release signing configuration. A GitHub runner has no
+keystore by default, so Gradle falls back to the debug key. The Play Console
+rejects those builds, and they fail to install over an existing copy of your app
+with `INSTALL_PARSE_FAILED_NO_CERTIFICATES`.
+
+`shorebird patch android` rebuilds your app to compare it against the release,
```
eseidel: I'm still not sure this is true.  Shorebird doesn't require it to my knowledge? But maybe gradle does

eseidel (commented):
This still feels like it could link to someone's action or well maintained guide rathr than us duplicating.

src/content/docs/code-push/ci/github.mdx
eseidel: You're right. I checked `AndroidArchiveDiffer`: `changedFiles` does report `META-INF/ANDROIDD.SF`, `META-INF/ANDROIDD.RSA` and `META-INF/MANIFEST.MF` as changed (asserted in `android_archive_differ_test.dart:58-60`), but the warnings come from `containsPotentiallyBreakingAssetDiffs` and `containsPotentiallyBreakingNativeDiffs`, which only look at `assets`/`res`/`AssetManifest.json` and `.dex`. `META-INF` matches neither, so a signature difference never triggers a warning.

Paragraph removed.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/content/docs/code-push/ci/github.mdx
eseidel: Reframed. It now opens with "`shorebird release android` and `shorebird patch android` run `flutter build`, so signing works exactly as it does for any Flutter app" — no "Shorebird does not sign your app".

---
_Generated by [Claude Code](https://claude.ai/code)_

src/content/docs/code-push/ci/github.mdx
eseidel: Cut it down. The section is now 40 lines instead of 90: a link to our [Releasing Flutter apps: Android](https://docs.shorebird.dev/flutter-concepts/releasing-flutter-apps/android/) guide for keystore creation and Gradle wiring, then the one CI-specific step (decode a base64 secret, write `key.properties`), what happens without it, and a one-line `.gitignore` warning. Gone: the separate secret-storage subsection, the `env:` vs interpolation note, and the full end-to-end workflow.

On whether even that much is needed: nothing here is Shorebird-specific, so it could be dropped entirely and left to Flutter's docs. I kept the one snippet because #338 asked for keystore instructions in this guide, and neither our Android guide nor Flutter's covers the CI half — both stop at local `key.properties` with a path to a keystore that exists on your machine. Happy to cut it to just the link if you'd rather.

I could not look at zap — `add_repo` returns "you don't have access to shorebirdtech/zap" for this session. If its workflow uses an action I do not know about, point me at the step and I will match it. I looked for a standard first-party Android signing action and did not find one that fits Flutter: the common community ones (for example `r0adkll/sign-android-release`) sign an already-built artifact, whereas Gradle signs during `flutter build`, so the keystore has to be in place before the build runs.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/content/docs/code-push/ci/github.mdx
eseidel: Removed — the full workflow example is gone, along with the secret-storage subsection and the `env:` note. Details on the standard-action question in [the thread above](https://github.com/shorebirdtech/docs/pull/664#discussion_r3996445445): I did not find one that works for Flutter, since Gradle signs during `flutter build` rather than after it, but I could not check zap.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/content/docs/code-push/ci/github.mdx
eseidel: It was Gradle, and my sentence overstated it. Shorebird requires nothing here — `flutter build` runs Gradle, and Gradle uses your release `signingConfig` if `key.properties` is present and the debug key if it is not.

The line now reads: "Without this, Gradle falls back to the debug key. The Play Console rejects those builds, and they fail to install over an existing copy of your app with `INSTALL_PARSE_FAILED_NO_CERTIFICATES`." No claim that Shorebird requires it.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/content/docs/code-push/ci/github.mdx
eseidel: Dropped the snippet. The section is now three short paragraphs: `flutter build` does the signing so it is standard Flutter, links to [Flutter's instructions](https://docs.flutter.dev/deployment/android#sign-the-app) and ours, then the one thing neither covers — on CI the keystore is not already on the machine, so store it as a base64 secret, decode it outside the checkout, and point `key.properties` at it. Then the consequence of not doing it.

17 lines, no YAML.

I saw the `_zap` release workflow (the repo is `_zap`, which is why `add_repo` failed on `zap`). Two things I took from it, less from the YAML than from the approach:

- It decodes to `$RUNNER_TEMP/upload.jks`, not into the checkout. My snippet wrote the keystore into `android/app/`, which was worse. The prose now says "outside your checkout".
- It writes no `key.properties` at all — it passes `ANDROID_KEYSTORE_PATH` and friends as env vars, which works because its `build.gradle` reads them. That is not what our own Android guide documents, so I did not generalize it into advice.

It also confirmed there is no standard action for this half: zap hand-rolls the decode step and uses `r0adkll/upload-google-play` only for the upload, after the build. Nothing signs before `flutter build` because Gradle does that during it.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/content/docs/code-push/ci/github.mdx
eseidel: Done — the guide is now links plus the CI-specific delta, with no YAML. Details in [the thread above](https://github.com/shorebirdtech/docs/pull/664#discussion_r3996530400).

On the standard action: there isn't one for this half. Signing happens inside `flutter build` via Gradle, so the keystore has to be in place beforehand; the common actions (`r0adkll/sign-android-release` and similar) sign an artifact after it is built. `_zap` hand-rolls the decode step for the same reason and only uses an action (`r0adkll/upload-google-play`) for the upload.

---
_Generated by [Claude Code](https://claude.ai/code)_

AbhishekDoshi26 (approved):

PR #669, opened 2026-09-16
docs: make it clear that deleting apps, releases, and patches is permanent

AbhishekDoshi26:
## Why

Deleting an app, a release, or a patch is permanent, but the docs only spelled
that out for *account* deletion. Users regularly reach out asking support to
restore a deleted app or release, and there is nothing to restore.

## What changed

- **New page: Delete Apps, Releases, and Patches** (`code-push/delete.mdx`, sidebar "Delete", lands between Tracks and Uninstall). One findable place to point users at: what each deletion removes, where the confirmation dialog appears, and the reversible alternatives.
- **`release.mdx`** - the existing danger callout said only "not reversible". Now titled "Deleting a release is permanent", states it is not restorable, notes the console confirmation dialog, and points to rollback. Added a line after the platform-selection screenshot noting that confirming is immediate and permanent.
- **`patch.mdx`** - new "Delete a patch" section under *Manage patches*. There is no way to delete an individual patch, so patches are only removed along with their release or app. The docs now say that instead of implying a patch can be deleted on its own.
- **`faq.mdx`** - new entry "Can you restore a deleted app, release, or patch?" under *Use cases & limitations*, since that is where people searching are likely to land.
- **`account/delete-account.mdx`** - cross-link to the new page.

Throughout, the docs steer users toward **rollback** rather than deletion for a
bad patch. Rollback is genuinely reversible (the console exposes a "Roll Forward"
action), so it is the right answer for the case that usually prompts someone to
delete a release in a panic.

## Notes for review

- An earlier draft made a claim about data not being recoverable that went
  further than it should have about how things work internally. I have cut it.
  The published text now says only what is true from the user's side: no trash,
  no archive, no undo, and support will not restore an individual deleted
  resource. Ping me offline for the detail behind that change.
- The docs say devices stop receiving patches for a deleted release version.
  I deliberately did **not** claim either way whether a device that already
  installed a patch keeps running it, since I could not confirm the device-side
  behavior. If you know the answer, it is worth stating explicitly.
- The app-deletion sentence about installed apps continuing to work is taken
  from the delete-app confirmation dialog's own copy.
- Follow-up PR documents the "Roll Forward" action, which `rollback.mdx`
  currently does not mention at all.

## Verification

`prettier`, `cspell`, and `vale` clean; `astro build` passes with "All internal
links are valid".


AbhishekDoshi26:
## Self-review

Re-read this with fresh eyes against the console and backend source. One real accuracy bug found and fixed in ba1b463; the rest is context for review.

### Fixed: the page described two different release deletions as one

The platform picker in the delete dialog is not just a scope filter, it selects a different operation:

- **All** deletes the release itself along with its patches.
- **A single platform** deletes only that platform's release and patch binaries. The release survives, and other platforms are untouched.

The page originally said deleting a release "deletes the release, its artifacts, and every patch published to it for the platforms you select," which is wrong for the single-platform case, and the follow-on claim that "devices will no longer receive patches for that release version" only holds for **All**. Both are now split out explicitly. Please sanity check that split against what you know of the intended behavior, since it is the one thing here I would most expect to have gotten subtly wrong.

### Verified

- The hard-delete claim. App and release deletion are unconditional `DELETE` statements with no soft-delete column, and per-platform deletion removes the objects from the release and patch artifact buckets. Nothing here is recoverable by the user.
- "Patches cannot be deleted on their own." There is no patch delete route, console component, or CLI command. Patches go away only with their release or app.
- Every deletion described does show a confirmation dialog, account deletion included.
- "A rollback can be undone" holds wherever rollback is offered at all, which is the same condition under which the console offers Rollback rather than Deactivate. #670 documents that.
- The app-deletion sentence about installed apps continuing to work is the delete-app dialog's own copy.

### Still open for you

- I did **not** claim whether a device that already installed a patch keeps running it after its release is deleted. The server stops advertising the patch, but the uninstall signal is the rolled-back-patch list, which would be empty for a deleted release, and the updater is in another repo so I could not confirm the device half. This is probably the single most common thing a user will want to know, so worth stating outright if you know it.
- An earlier draft asserted something stronger than it should have about internal recovery mechanisms. That wording is gone; the text now speaks only to what the user can and cannot do. Ping me offline for the detail.
- `order: 8` ties with `uninstall.mdx`. It resolves to Tracks, Delete, Uninstall, which reads fine, but say the word if you want an explicit order.

`prettier`, `cspell`, `vale` clean; `astro build` reports all internal links valid.

src/content/docs/code-push/delete.mdx
```diff
@@ -0,0 +1,103 @@
...
+  What deleting an app, release, or patch in Shorebird removes, and the options
+  to consider first.
+sidebar:
+  label: Delete
+  order: 8
+---
+
+import { LinkCard } from 'starlight-theme-nova/components';
+
+:::note[Deleting is permanent]
```
dawn-ducky: Deleting is permanent and cannot be undone. Shorebird is unable to restore deleted apps, releases or patches once deleted.

src/content/docs/code-push/delete.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+Once an app, release, or patch is deleted, it can't be restored, including by
+Shorebird support. If you're only trying to stop a patch or hand off an app, see
```
dawn-ducky: Change this to why a person would want to delete an app, patch or release. Talk about rollbacks instead of deleting. This seems redundant with the note above. 

This seems to be covered in the before you delete section. That should be moved up to the top.

src/content/docs/code-push/delete.mdx
```diff
@@ -0,0 +1,103 @@
...
+[Before you delete](#before-you-delete) for options that can be undone.
+
+:::
+
+The Shorebird console shows a confirmation dialog before any of these deletions,
+so you can review what's about to be removed before anything happens.
+
+## What each deletion removes
```
dawn-ducky: This section is confusing. I'm not sure what it is trying to communicate to the user. the sections below are much clearer.

src/content/docs/code-push/delete.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+<LinkCard
+  title="Roll back a Patch"
+  href="/code-push/rollback"
+  description="Remotely uninstall a patch. You can undo a rollback later."
+/>
+
+## Delete a release
+
+The delete dialog asks which platforms to delete, and the answer changes what
```
dawn-ducky: Releases are deleted per platform. Depending on whether you choose all or a specific platform impacts what is deleted.

src/content/docs/code-push/delete.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## Delete an app
+
+1. Navigate to the app in the
+   [Shorebird console](https://console.shorebird.dev/).
+2. Go to the **Settings** tab.
+3. In the **Danger Zone**, select **Delete**.
+4. Confirm in the dialog that appears.
+
+Deleting an app also deletes every release, patch, and artifact associated with
```
dawn-ducky: Move this before the steps on how to delete an app.

src/content/docs/code-push/delete.mdx
```diff
@@ -0,0 +1,103 @@
...
+it. Copies of the app that are already installed on end users' devices keep
+working normally, but they won't receive new patches.
+
+If you want to hand an app to someone else, use **Transfer Ownership** in the
```
dawn-ducky: If you want to move an app to another organization, ...

dawn-ducky (changes requested):
Please make changes for improved clarity.

src/content/docs/code-push/delete.mdx
AbhishekDoshi26: Used this wording for the note in 02499a9.

src/content/docs/code-push/delete.mdx
AbhishekDoshi26: Removed the redundant sentence from the note and moved "Before you delete" to the top in 02499a9. It now opens with why someone might want to delete, and leads with rolling back a bad patch instead of deleting its release.

src/content/docs/code-push/delete.mdx
AbhishekDoshi26: Agreed, removed the table in 02499a9. The sections below cover the same ground more clearly.

src/content/docs/code-push/delete.mdx
AbhishekDoshi26: Reworded in 02499a9: the section now opens with "Releases are deleted per platform" and explains that choosing All or a single platform determines what is removed.

src/content/docs/code-push/delete.mdx
AbhishekDoshi26: Moved the paragraph above the steps in 02499a9.

src/content/docs/code-push/delete.mdx
AbhishekDoshi26: Updated in 02499a9 to "If you want to move an app to another organization, use **Transfer Ownership** in the same Danger Zone instead of deleting it." I applied the same framing to the matching bullet in Before you delete.

PR #670, opened 2026-09-16
docs: document the Roll Forward patch action

AbhishekDoshi26:
## Why

The console has a **Roll Forward** action on rolled back patches, but
`rollback.mdx` never mentioned it. The docs read as though rolling back were a
one-way door, which is the opposite of the point we make elsewhere when we tell
users to roll back rather than delete a release.

## What changed

All in `rollback.mdx`:

- New **"Roll a patch forward"** section: where the action lives in the console, that it only appears on a patch that is *currently* rolled back, and what happens on device (the patch stops being reported as rolled back, so devices pick it up on their next patch check).
- A note that devices downgraded by the rollback have to download the patch again and that it counts against monthly patch installs, mirroring the caveat already documented for rollback itself.
- A **"When rolling forward is unavailable"** subsection: releases built on a Flutter version without rollback support offer **Deactivate** instead, which behaves differently on device (already-installed devices keep the patch) and is undone by moving the patch back to the `stable` track rather than by rolling forward.
- One line in the intro pointing out that a rollback is reversible.
- A console screenshot of the **Roll Forward** menu item on a rolled back patch (`src/assets/roll_forward_button.png`), matching the existing screenshot for the Rollback action.

## Verification

`prettier`, `cspell`, and `vale` clean; `astro build` passes with "All internal
links are valid", including the new `#roll-a-patch-forward` anchor.


AbhishekDoshi26:
## Self-review

Re-read this against the console source and the rest of the docs. Two wording fixes pushed in 996608e, plus one correction I had already made before opening the PR.

### Fixed: update timing did not match our own FAQ

I wrote that devices "pick it up on their next patch check and install it the next time the app starts." Per "When do updates happen?", the updater checks on startup, downloads in the background during that session, and *applies* the patch on the next restart. Now worded as download on the next check, apply on the next restart, so the two pages agree.

### Fixed: overstated when Rollback appears

"A patch that has never been rolled back shows Rollback instead" is only true on releases that support rollback at all. On releases that do not, the menu offers Deactivate and neither Rollback nor Roll Forward appears. Reworded to say the same menu offers Rollback on a patch that is not rolled back, which is true in the context where Roll Forward is relevant.

### Corrected before opening: deactivation is reversible

My first draft said deactivation could not be undone and that you would have to publish a new patch. That is wrong. Deactivate promotes the patch to a `deactivated` channel, and Change Track is always available, so moving it back to `stable` reverses it. The section says that now and gives the `patches set-track` equivalent. This is the claim I would most like a second opinion on, since I am describing the mechanism rather than repeating existing copy.

### Verified

- Roll Forward appears only when the patch is currently rolled back, and only on releases that support rollback.
- Rolling forward clears the rolled-back flag and the cached patch-check response, so the patch becomes servable again, matching "becomes the active patch for its track."
- The patch-install caveat follows from the billing page's own definition (a successful update applied on a device) and mirrors the caveat already documented for rollback itself.

### Still open for you

- The page's existing note that rollback "requires a minimum Flutter version of 3.27.4" does not match how the check actually works. Support is decided by a revision blocklist with no version floor, because the feature was backported across fork branches, so a strict reading of that sentence is not the real gate. I left it alone as out of scope. Happy to reword it here or separately.
- No screenshot for the Roll Forward menu item, while the rollback section has one for its button. Point me at an image and I will add it.

`prettier`, `cspell`, `vale` clean; `astro build` reports all internal links valid, including the new `#roll-a-patch-forward` anchor.

src/content/docs/code-push/rollback.mdx
```diff
@@ -1,6 +1,7 @@
 ---
 title: Roll back a Patch
-description: Learn how to roll back a patch in Shorebird.
+description:
+  Learn how to roll back a patch in Shorebird, and how to roll it forward again.
```
dawn-ducky: Learn how to roll back and roll forward patches in Shorebird.

src/content/docs/code-push/rollback.mdx
```diff
@@ -124,6 +130,58 @@ remaining patch installs, the app will revert to the base release.
...
+
+### What happens when a patch is rolled forward?
+
+The patch stops being reported as rolled back, so it becomes the active patch
+for its track again. Devices download it on their next update check and apply it
+the next time the app restarts, exactly as they would a newly published patch.
+
+:::note
+
+Devices that were downgraded by the rollback have to download the patch again,
```
dawn-ducky, line 162: Remove "just as the original rollback did" This makes it sound like we charge when a patch is rolled back.

src/content/docs/code-push/rollback.mdx
```diff
@@ -124,6 +130,58 @@ remaining patch installs, the app will revert to the base release.
...
+and that counts against your monthly patch installs, just as the original
+rollback did.
+
+:::
+
+### When rolling forward is unavailable
+
+Roll Forward is available wherever rollback is. Releases built with a Flutter
+version that does not support rollback offer **Deactivate** instead of
```
dawn-ducky: What Flutter versions does this apply to?

dawn-ducky (changes requested):
Changes and questions for clarity

src/content/docs/code-push/rollback.mdx
AbhishekDoshi26: Updated the description to "Learn how to roll back and roll forward patches in Shorebird." in cec5d0b.

src/content/docs/code-push/rollback.mdx
AbhishekDoshi26: Good catch, removed in cec5d0b. The note now just says that devices downgraded by the rollback have to download the patch again, and that counts against monthly patch installs.

src/content/docs/code-push/rollback.mdx
AbhishekDoshi26: Roll Forward is available on any release that supports rollback, so the page's existing minimum version note (3.27.4) covers it. I first tried to state an exact cutoff here, but there isn't a clean one: Shorebird shipped more than one 3.22.3 build, and the earliest predates rollback support while the later ones include it. Rather than document that, I removed the section about releases that don't support rollback (and the Deactivate details) in eec4aee.

PR #673, opened 2026-09-16
chore: update Flutter to 3.47.4

nickshorebird:
## Summary

Shorebird 1.6.122 shipped Flutter 3.47.4 support today, so the docs advertise a version users can no longer get as the newest. Dart stays at 3.13.3, since upstream 3.47.3 and 3.47.4 pin the same dart_revision at 1d1a730ef91.

## Testing

- Verified against the shipped release: flutter_release/3.47.4 resolves in all three repos and stable is at 1.6.122.


nickshorebird:
Force merging. Docs update apart of engine release.

PR #675, opened 2026-09-17
docs: document Shorebird with long-running Android background processes (shorebirdtech/shorebird#2396)

AzazelSensei:
## Status

**READY**

## Description

I added a FAQ entry and a troubleshooting note for Android apps that keep a foreground service running. Swiping the app from Recents does not kill the process, so the patch never applies until a real process restart.

See shorebirdtech/shorebird#2396


src/content/docs/code-push/faq.mdx
```diff
@@ -545,6 +545,29 @@ push notification.
...
+keeps the process alive after the user swipes the app out of the recent-apps
+list. That swipe is not a process restart, so bringing the UI back still runs
+the previous Dart code.
+
+To receive patches while the process stays alive:
+
+1. Check for and download patches while the process is running with
+   [package:shorebird_code_push](https://pub.dev/packages/shorebird_code_push).
+   Default automatic updates only run on process start.
+2. Apply the patch with a real process restart: Settings → Apps → Force stop,
```
AbhishekDoshi26: This is the one substantive change I would ask for.

`package:shorebird_code_push` has API built for precisely this situation:

```dart
enum UpdateStatus {
  /// The app is up to date, but a restart is required for the update to take effect.
  restartRequired,
  ...
}

Future<Patch?> readCurrentPatch();
Future<Patch?> readNextPatch();  // the most recently downloaded patch
```

`UpdateStatus.restartRequired` means "a patch is downloaded and staged, waiting for a restart". That is the state a foreground-service app will sit in indefinitely, and it is detectable from Dart.

As written, step 1 uses the package and step 2 drops it and tells the reader to force stop from Android settings. That inverts the usefulness: force stop is a developer testing technique, not something a shipping app can ask its users to do, and a user who waits "until the OS kills the process" may wait days.

Suggested rewrite of the list:

```mdx
To receive patches while the process stays alive:

1. Check for and download patches while the process is running with
   [package:shorebird_code_push](https://pub.dev/packages/shorebird_code_push).
   Default automatic updates only run on process start.
2. Detect the staged patch. `checkForUpdate()` returns
   `UpdateStatus.restartRequired` once a patch is downloaded and waiting, and
   `readNextPatch()` returns it. Use that to prompt the user to restart at a
   point that suits your app.
3. Apply the patch with a real process restart. Stopping the foreground service
   and then swiping the app away is enough, as is waiting until the OS kills the
   process.

While testing, you can force a restart with **Settings → Apps → Force stop**.
```

That keeps everything you already had, puts the package in charge of the part it can actually automate, and demotes force stop to the testing note it really is.

src/content/docs/code-push/troubleshooting.mdx
```diff
@@ -98,6 +98,14 @@ Sending patch check request: PatchCheckRequest {
 
 Only patches created for this release version will be compatible with your app.
 
+**You swiped the app away on Android but the patch did not apply.**
```
AbhishekDoshi26: Phrasing: the other causes under **Patch not showing up** are written as causes, this one is written as a symptom.

```
44: **The app is running in Debug or Profile mode.**
53: **You are testing on an iOS Simulator.**
62: **The patch was created for a different release version than the one running on your device/emulator.**
```

Each completes "the patch is not showing up because ...". Yours reads as a restatement of the section heading instead.

**Suggested:**

```mdx
**The app's process is still running because of a foreground service.**
```

Optional, take it or leave it: the third cause above yours adds **How to tell if this is the problem** and **How to fix it** sub-headings. The first two causes do not, so it is not a hard convention, but this failure is genuinely hard to self-diagnose, so a "how to tell" line would earn its place. Something like: the patch check request appears in the logs but `Shorebird updater: no active patch` persists across a swipe-away, and `adb shell ps | grep <your package>` still shows the process.

src/content/docs/code-push/faq.mdx
```diff
@@ -545,6 +545,29 @@ push notification.
 See [Update Strategies](/code-push/update-strategies) for more information about
 how to configure this behavior.
 
+### How do you use Shorebird with a long-running Android background process?
+
+Shorebird applies a downloaded patch the next time the process starts, not when
```
AbhishekDoshi26, line 552: Good precision here, and it is the sentence that makes the whole entry worth having.

Flagging one thing for you rather than asking for a change: this directly contradicts **When do updates happen?** twelve lines above, which says updates are "applied the next time the app is restarted". Yours is the accurate version. A reader who finds the earlier entry first walks away with the imprecise one.

If you are willing, tightening line 537 to say process restart, and linking down to this entry, would make the page self consistent. Entirely reasonable to say that is out of scope for this PR, in which case it is worth a follow up issue so it does not get lost.

AbhishekDoshi26 (changes requested):
Thanks for picking this up, and for tying it to a concrete plugin rather than describing the problem in the abstract. This lands in the right two places, the FAQ heading matches the house "How do you ..." phrasing, and it's slotted next to **When do updates happen?**, which is where a reader would look. I checked the technical claims against the updater source and the maintainers' own statements and they hold up, which is not a given for a subtle topic like this.

I built the branch and ran the full lint suite. Two things block merge, one of which is not your fault, and there are a few structural points worth addressing before this goes in.

## Blocking

### 1. `prettier` fails, which is what is making the `build` check red

Two lines in `faq.mdx` are wrapped a column short:

```
565:    stop the foreground service and then swipe the app away, or wait until the
566:    OS kills the process.
568: Recreating the Flutter engine without killing the process is not a supported
569: way to apply a patch.
```

Run `npm run format` and commit the result. That is the whole fix, nothing else in the repo is touched.

### 2. `style-check` is red, but not because of this PR

Ignore it. Vale 3.22.0 against your branch reports exactly one error, in `flutter-concepts/how-to-install-flutter.mdx:342`, a file you did not touch. It is a `main` branch breakage caused by Vale shipping a new release, and shorebirdtech/docs#678 fixes it. Once that merges, re-running this job will go green with no change from you.

## Worth fixing before merge

### 3. Step 2 is developer instructions, not user instructions

This is the main substantive point, and it's in the inline comment on line 564.

Short version: `package:shorebird_code_push` exposes `UpdateStatus.restartRequired` and `readNextPatch()`, which exist for exactly this case. The FAQ currently spends step 1 on the package and then falls back to "Settings, Apps, Force stop" for step 2. Force stop is a fine way for *you* to verify behavior, but it is not something an app can ask its users to do, and this FAQ is aimed at shipping apps.

### 4. It overlaps with the entry 12 lines above it

**When do updates happen?** (line 533) already says updates "will be applied the next time the app is restarted" and already links the same package and the same Update Strategies page. Your entry is more precise, which is good, but it means a reader who hits the earlier entry first gets the vaguer answer and may never scroll to yours.

Consider sharpening the earlier entry's "the next time the app is restarted" to say process restart, so the two agree, and let your entry focus on what is specific to background processes. Either that or a short pointer from the earlier entry down to yours.

### 5. The troubleshooting bullet is phrased as a symptom, its siblings are causes

Detail in the inline comment on line 101.

## Verified, no action needed

I want to call these out because they are the parts that are easy to get wrong:

- **"Recreating the Flutter engine without killing the process is not a supported way to apply a patch"** is correct, and matches eseidel's own reasoning in shorebirdtech/shorebird#2350: *"The reason why we don't offer 'hot restart' of sorts is that we can't know if plugins will manage the restart correctly."* Note that the reporter there confirmed engine recreation does work on Android in practice, so "not supported" is exactly the right framing, rather than "does not work".
- **"not when an Activity is recreated"** is right. An Activity recreation reuses the engine, so no patch is picked up.
- **"Default automatic updates only run on process start"** matches the Update Strategies page.
- **The cross-link anchor resolves.** I built the site and confirmed `id="how-do-you-use-shorebird-with-a-long-running-android-background-process"` is emitted, and the internal link validator passes on all pages.
- **The link style** `[package:shorebird_code_push](...)` without backticks matches the adjacent FAQ entry at line 541. `update-strategies.mdx` uses backticks, so the repo is inconsistent here, and matching your neighbour is the right call.
- cspell passes, and the build succeeds with 74 pages.

## Nit

Line 564: the repo bolds UI navigation paths, for example `**Account → API Keys**` in this same file at line 73, and `**Settings → Privacy & Security → For Developers**` in `how-to-install-flutter.mdx:195`. Yours is unbolded.

src/content/docs/code-push/faq.mdx
AzazelSensei: Rewrote the list around restartRequired and readNextPatch, and moved Force stop down to a testing note.

src/content/docs/code-push/troubleshooting.mdx
AzazelSensei: Heading is a cause now. Added the log line plus adb shell ps as the check.

src/content/docs/code-push/faq.mdx
AzazelSensei: Tightened that entry to process restart and linked it down here.

AzazelSensei:
Pushed the review notes: package API for the staged patch, Force stop as a testing note, process restart in When do updates happen, troubleshooting heading as a cause, and prettier wrap.

src/content/docs/code-push/troubleshooting.mdx
```diff
@@ -98,6 +98,19 @@ Sending patch check request: PatchCheckRequest {
 
 Only patches created for this release version will be compatible with your app.
 
+**The app's process is still running because of a foreground service.**
+
+If the app runs a foreground service or other long-running background work,
+swiping it from the recent-apps list does not kill the process. Shorebird loads
+the patch on the next process start.
+
+A patch check request in the logs with `Shorebird updater: no active patch`
```
AbhishekDoshi26: I gave you this wording last time and it is wrong twice over. Apologies for the churn.

**1. `Shorebird updater: no active patch` does not distinguish this cause.**

Line 80, 27 lines above, already uses that exact string as the diagnostic for a completely different cause:

```
80: You will see `Shorebird updater: no active patch` in your device logs.
```

That is the tell for "the patch was created for a different release version". Using it again here means the page now offers the same log line as the distinguishing symptom for two different problems, so it distinguishes neither.

It is also just inaccurate when the app already has a patch applied. In that case the log reports the active patch number, not "no active patch", even though the process is still alive and the newer patch is stuck waiting. The reader would conclude this is not their problem when it is.

**2. `adb shell ps | grep <your package>` does not work on modern Android.**

Since the switch to toybox, `ps` with no arguments only lists processes owned by the calling user, which for `adb shell` is `shell`, not your app. It will print nothing whether or not your app is running, which points the reader at the wrong conclusion. It needs `ps -A`, and `pidof` is cleaner anyway.

**Suggested replacement for lines 107 to 109:**

```mdx
**How to tell if this is the problem**

The process ID does not change when you swipe the app away. Run
`adb shell pidof <your.package.name>` before and after swiping: if it prints the
same PID both times, the process never died and no patch will be applied.

From Dart, the equivalent signal is `checkForUpdate()` returning
`UpdateStatus.restartRequired`, or `readNextPatch()` reporting a higher patch
number than `readCurrentPatch()`.
```

Both of those are unambiguous, neither collides with the diagnostic for the version-mismatch cause above, and the second ties back to the API the FAQ entry now recommends.

I also switched it to the `**How to tell if this is the problem**` sub-heading, matching the cause directly above at line 78. Optional, but since this cause now has a real diagnostic it reads better in the same shape as its neighbour.

AbhishekDoshi26 (changes requested):
Re-reviewed the update. You addressed all five points, and CI is now fully green (`build`, `spell-check`, `style-check` all pass, since #678 landed and cleared the unrelated Vale failure).

The rewritten step 2/3 split is exactly right, and sharpening **When do updates happen?** to say "the next time the process restarts" with a link down to the new entry fixes the contradiction properly rather than papering over it. Nice.

**One thing left, and it is my mistake, not yours.** The "how to tell" diagnostic I suggested in my last review is wrong on two counts. You implemented what I wrote, so this is me correcting my own bad advice. Details in the inline comment on line 107, corrected replacement included.

Everything else is good to merge once that is sorted.

## Verified on this revision

- prettier, cspell, Vale 3.22.0, and the component-label linter all clean.
- `astro build` succeeds, 74 pages, internal link validator passes.
- Both new cross-links resolve: the same-page relative `#how-do-you-use-...` in `faq.mdx` and the absolute `/code-push/faq/#how-do-you-use-...` in `troubleshooting.mdx`. Using a relative anchor for the same-page link and an absolute one for the cross-page link is the correct call.
- The cause line now matches its siblings under **Patch not showing up**, and `**Settings → Apps → Force stop**` is bolded like the other UI paths in the repo.
- The `UpdateStatus.restartRequired` and `readNextPatch()` references match the current `shorebird_code_push` 2.0.7 API in `shorebirdtech/updater`.

src/content/docs/code-push/troubleshooting.mdx
AzazelSensei: Switched that check to pidof plus restartRequired/readNextPatch, and put it under How to tell / How to fix like the cause above.

AbhishekDoshi26 (approved):
This is ready. You fixed the diagnostic, and you went one better by adding the `**How to fix it**` sub-heading as well, which I had not asked for but which makes this cause match the structure of the release-version cause directly above it. The troubleshooting entry now reads exactly like its siblings.

Thanks for working through three rounds on this, including one where the change I asked for was my own bad suggestion. The end result is better than what I originally proposed.

## Verified on this revision

Ran the full suite against the merge of your branch with current `main`, which is what CI actually tests:

- **Vale 3.22.0**: 0 errors across 74 files.
- **prettier**, **cspell**, **component-label linter**: all clean.
- **`astro build`**: succeeds, 74 pages, internal link validator passes.
- **Both cross-links resolve.** `id="how-do-you-use-shorebird-with-a-long-running-android-background-process"` is emitted in the built FAQ page, and the troubleshooting page links to it.
- CI on the PR is green on `build`, `spell-check` and `style-check`.

Your branch is one commit behind `main` (#678 landed after you pushed), but since the checks run against the merge result and that is green, there is nothing for you to do about it.

## Technical claims, re-checked against source

- `UpdateStatus.restartRequired` and `readNextPatch()` match the current `shorebird_code_push` 2.0.7 API in `shorebirdtech/updater`, including the semantics you describe: `readNextPatch()` returns the most recently downloaded patch and matches `readCurrentPatch()` when nothing new has been downloaded.
- `adb shell pidof <your.package.name>` is the right call over `ps`, and comparing the PID across the swipe is unambiguous.
- "Recreating the Flutter engine without killing the process is not a supported way to apply a patch" matches eseidel's reasoning in shorebirdtech/shorebird#2350.
- "not when an Activity is recreated" is correct, since an Activity recreation reuses the engine.
- Sharpening **When do updates happen?** to "the next time the process restarts" removes the contradiction at its source rather than leaving two answers on one page.

This resolves shorebirdtech/shorebird#2396. Approving.

PR #679, opened 2026-09-18
Add changelog page

dawn-ducky:
## Status

**READY**

## Description

Adds `/changelog/` — a searchable, faceted feed of what shipped to Code Push, the CLI, and the API, built as a `StarlightPage` next to `roadmap.astro` and linked from the sidebar right below it.

Each entry carries a type badge (New/Fixed/Changed/Deprecated), an area tag, and a version, and expands in place to show detail bullets, an optional command snippet, a link to the relevant doc page, and a copyable permalink (`/changelog/#entry-id`). Opening the page with a matching `#entry-id` in the URL auto-expands and scrolls to that entry. Search and the area chips filter the list client-side.

Seeded with a single example entry (staged rollouts) rather than the full sample dataset from the design exploration — add more to the `ENTRIES` array in `src/pages/changelog.astro` as things ship.

Design source: https://claude.ai/design/p/8d6abb36-5f0f-4da2-86d9-636e825acb58 ("Docs Changelog.dc.html", option 3a — the full-page direction marked ready to ship).

## Test plan

- [x] `npm run build` (`astro check && astro build`) passes clean: 0 errors, 0 warnings, all internal links valid
- [x] Verified in a local preview: entry expands/collapses, search filters to zero results and shows the empty state, area chip filtering, light/dark theme both render correctly matching the rest of the site


easymac:
I think this looks great. I made a couple tiny adjustments:
image

- Vertically centered the filter chips
- Vertically centered the + icon for expanding an entry, and made it a little bigger
- Made it so you can click anywhere in the title to expand the summary

The last change isn't precious, just an idea :)

AbhishekDoshi26:
Reviewed this. The page structure, the filtering, and the sidebar wiring all look good, and the build is clean (0 errors, 0 warnings). I found a few small issues and pushed the fixes as a55b990:

1. **The copy-link fallback never worked.** The `.cl-copy` handler called `ev.preventDefault()` before writing to the clipboard. So when the clipboard write failed (permissions, insecure context), nothing happened, even though the comment says "the link itself still works as a fallback". The click also gave no sign that it had copied anything. Now the link navigates normally, so the permalink always ends up in the address bar, and a successful copy briefly shows ✓ with the aria-label "Link copied".
2. **Permalinks only opened on first load.** The auto-expand ran once, in `init()`. Following a `#entry-id` link from the page itself (or pasting one into the same tab) only fires `hashchange`, so the entry stayed collapsed. The same logic now runs on `hashchange` too.
3. **Search skipped most of an entry.** `data-search` held only the title, summary, area, and type. Searching a version number or a flag that appears only in the details returned nothing. It now also covers the version, the bullets, and the command.
4. **The command had a stray leading space.** `<pre>` preserves whitespace, so the template's indentation before `<code>` rendered as a leading space ahead of `$`. They're on one line now.

How I checked: ran `npm run build` and cspell. In a local preview I confirmed that a hash change expands the entry, that version and bullet searches match, that the empty state still appears, and that a real click on copy writes the full URL, shows ✓, and reverts after 1.5s.

AbhishekDoshi26:
One more fix, pushed as 3cf311f: the changelog wasn't reachable as Markdown for agents.

`Head.astro` gives every page a `<link rel="alternate" type="text/markdown" href="<path>.md">`, but `[...slug].md.ts` only builds `.md` files for content collection pages. `changelog.astro` is a standalone page, so the page advertised `/changelog.md`, which would 404, and a request for `/changelog/` with `Accept: text/markdown` quietly got HTML instead.

- Moved `ENTRIES`, `AREAS`, the entry type, and the month grouping into `src/data/changelog.ts`, so the page and the Markdown route read the same list. You still add entries in one place, just in that file now.
- Added `src/pages/changelog.md.ts`, which renders `/changelog.md`: frontmatter, month headings, and for each entry the type, area, version, date, permalink, bullets, command, and doc link. The existing middleware now negotiates `/changelog/` to it with no changes.
- Listed `/changelog.md` under Optional in `llms.txt`.

How I checked: `npm run build` is clean (0 errors, 0 warnings), and Prettier and cspell pass on the changed files. Using `wrangler pages dev` with the real middleware:
- `Accept: text/markdown` on `/changelog/` returns 200 `text/markdown` with `Vary: Accept` and CORS.
- A browser `Accept` header still gets HTML.
- `text/markdown;q=0.5, text/html` still gets HTML.
- `/changelog.md` returns 200.

AbhishekDoshi26:
Pushed 3d897b3 to fix the type badge colors.

**New looked like the area badge.** It used `--sl-color-accent`, and the Nova theme's accent is gray, so the NEW pill had almost the same gray text and fill as the CODE PUSH pill next to it. Each type now has its own color: **New is green**, and **Fixed moves from green to blue** so the two don't clash. That matches the usual changelog convention.

**Light-mode contrast.** These are about 11px bold labels, so they need 4.5:1. Measured against the tinted pill on a white page:

| Type | Before | After |
|---|---|---|
| New | gray, same as the area badge | `#166534`, 6.2:1 |
| Fixed | `#16a34a`, 2.9:1 | `#1d4ed8`, 5.6:1 |
| Changed | `#9333ea`, 4.4:1 | `#7e22ce`, 5.7:1 |
| Deprecated | `#d97706`, 2.8:1 | `#92400e`, 6.3:1 |

Dark mode already passed. New uses `#4ade80` (8.4:1) and Fixed uses `#93c5fd` (8.7:1); Changed and Deprecated keep their existing shades.

How I checked: `npm run build` is clean (0 errors, 0 warnings). In a local preview I added the other three badge types to the page and confirmed the computed colors and the look in both light and dark themes.

src/data/changelog.ts
```diff
@@ -0,0 +1,76 @@
...
+  docLink?: { label: string; href: string };
+}
+
+// Hand-written, newest first. Add an entry here when something ships that a
+// developer integrating Code Push would want to know about.
+export const ENTRIES: ChangelogEntry[] = [
+  {
+    id: 'staged-rollouts-in-one-command',
+    date: '2026-09-10',
+    version: '1.7.2',
```
AbhishekDoshi26: **Blocker.** This entry describes a CLI change that hasn't shipped. The latest CLI release is v1.6.123 (2026-09-21), and there is no 1.7.x. `shorebird patch` has no `--rollout` flag and there's no `shorebird patch rollout set` subcommand. The linked guide, Percentage-based rollouts, describes building rollouts yourself with a KV store, not a CLI flag.

This page goes live on docs.shorebird.dev and is listed in `llms.txt`, so both readers and agents will take it as a real release note. Could you swap in a real recent change before merge? If there isn't one worth listing yet, the page can ship with an empty `ENTRIES`, since the empty state already handles that.

src/data/changelog.ts
```diff
@@ -0,0 +1,76 @@
...
+function formatMonth(iso: string): string {
+  return new Date(iso).toLocaleDateString('en-US', {
+    month: 'long',
+    year: 'numeric',
+    timeZone: 'UTC',
+  });
+}
+
+/** `ENTRIES` grouped under "September 2026"-style headings, newest first. */
+export function groupByMonth(
```
AbhishekDoshi26: **(fixing)** `groupByMonth` relies on `ENTRIES` being hand-sorted newest first. One entry out of place starts a second heading for its month. I'm sorting by `date` here instead, so the order in the array stops mattering.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+              .toLowerCase();
+            return (
+              <article
+                class="cl-entry"
+                id={e.id}
+                data-area={e.area}
+                data-search={searchText}
+              >
+                <div class="cl-head">
+                  <button
```
AbhishekDoshi26: **(fixing)** The button contains only the badges, so its accessible name is "NEW CODE PUSH 1.7.2" and the title isn't part of it. With more entries, a screen reader user hears a list of near-identical buttons. Fix: give the `h3` an id and set `aria-labelledby="<id>-title <id>-toggle"` on the button, so it reads the title followed by the badges.

AbhishekDoshi26 (commented):
Second pass on this, covering the code and the design in a local preview (dark and light).

**One blocker before merge: the seeded entry describes a feature that doesn't exist.** See the inline comment on `src/data/changelog.ts`. That one needs you, since it's about what we actually shipped.

Everything else is small. I'm pushing a commit right after this review that fixes the comments marked **(fixing)**:

- Accessible name of the entry toggle
- Title and summary text can't be selected
- Result count not announced
- Find-in-page can't reach collapsed details
- Month grouping depends on array order
- Double border on the command block
- Detail panel has no fill in light mode
- Commands and flags render as plain prose
- `type="search"` and `white-space` on the command

The comments marked **(suggestion)** are design ideas I left for you to decide on. Also, the PR description still says to add entries in `src/pages/changelog.astro`. Since 3cf311f they live in `src/data/changelog.ts`.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+    width: 100%;
+    padding: 0;
+    border: 0;
+    background: transparent;
+    cursor: pointer;
+    text-align: left;
+    outline: 0;
+  }
+
+  .cl-entry-toggle::after {
```
AbhishekDoshi26: **(fixing)** The `::after` overlay covers the whole header, so you can't select the title or summary text. A drag toggles the entry instead, and `elementFromPoint` on the summary returns the button. I'm replacing the overlay with a click handler on `.cl-head` that ignores links and clicks that end a text selection. Keyboard behavior stays on the button.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+      >
+      {
+        AREAS.map((a) => (
+          <button class="cl-chip" data-area={a} type="button">
+            {a}
+          </button>
+        ))
+      }
+    </div>
+    <span class="cl-count" id="cl-count">
```
AbhishekDoshi26: **(fixing)** Filtering updates this count silently, so screen reader users don't learn how many results matched. Adding `aria-live="polite"`, and writing the text only when it changes so it isn't announced on load.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+                          stroke-width="2"
+                          stroke-linecap="round"
+                        />
+                      </svg>
+                    </span>
+                  </button>
+                  <h3 class="cl-title">{e.title}</h3>
+                  <p class="cl-summary">{e.summary}</p>
+                </div>
+                <div class="cl-detail" id={`${e.id}-detail`} hidden>
```
AbhishekDoshi26: **(fixing)** With plain `hidden`, the browser's find-in-page can't match text in a collapsed entry (the bullets, the command). `hidden="until-found"` plus a `beforematch` listener that opens the entry fixes this. Browsers without support treat it as plain `hidden`. The CSS drops the panel's margin and padding while it's hidden, since `until-found` keeps the box rendered.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+                  <ul class="cl-bullets">
+                    {e.bullets.map((b) => (
+                      <li>{b}</li>
+                    ))}
+                  </ul>
+                  {e.code && (
+                    <pre class="cl-code"><code>$ {e.code}</code></pre>
```
AbhishekDoshi26: **(fixing)** Starlight's inline-code style (1px border, padding) also applies to this `<code>`, so the command gets a second box inside the block, in both themes. Resetting border, padding, and background on `.cl-code code`.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+    color: var(--sl-color-gray-2);
+  }
+
+  .cl-code {
+    margin: 1rem 0 0;
+    padding: 0.85rem 1.1rem;
+    border-radius: 0.75rem;
+    background: var(--sl-color-bg);
+    overflow-x: auto;
+    white-space: nowrap;
```
AbhishekDoshi26: **(fixing)** `nowrap` collapses newlines, so a multi-line command would render on one line. `pre` keeps them and still scrolls horizontally.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+    margin: 0.4rem 0 0;
+    max-width: 68ch;
+    font-size: 0.9rem;
+    line-height: 1.6;
+    color: var(--sl-color-gray-2);
+  }
+
+  .cl-detail {
+    margin-top: 1rem;
+    background: var(--sl-color-bg-sidebar);
```
AbhishekDoshi26: **(fixing)** `--sl-color-bg-sidebar` is white in light mode, the same as the page, so the expanded panel has no visible fill there. Using `--sl-color-gray-7` in light mode, and keeping the sidebar color in dark.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+                          d="M8 2v12"
+                          stroke="currentColor"
+                          stroke-width="2"
+                          stroke-linecap="round"
+                        />
+                      </svg>
+                    </span>
+                  </button>
+                  <h3 class="cl-title">{e.title}</h3>
+                  <p class="cl-summary">{e.summary}</p>
```
AbhishekDoshi26: **(fixing)** Flags and commands in the summary and bullets render as ordinary prose. I'm adding Markdown-style backticks: text in backticks renders as `<code>`, the Markdown twin gets it for free, and search strips the backticks.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+        ></circle>
+        <path
+          d="M21 21l-4.3-4.3"
+          stroke="currentColor"
+          stroke-width="2"
+          stroke-linecap="round"></path>
+      </svg>
+      <input
+        id="cl-search"
+        type="text"
```
AbhishekDoshi26: **(fixing)** `type="search"` gives a clear button and Esc-to-clear for free.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+      {ENTRIES.length}
+      {ENTRIES.length === 1 ? 'entry' : 'entries'}
+    </span>
+  </div>
+
+  <div class="cl-list" id="cl-list">
+    {
+      groups.map((group) => (
+        <div class="cl-group">
+          <h2 class="cl-month">{group.label}</h2>
```
AbhishekDoshi26: **(suggestion)** "On this page" only shows "Overview" because `StarlightPage` gets no `headings`. If these month `h2`s got ids (for example `2026-09`) and were passed as `headings`, the right-hand panel would become a date index once there are more entries.

src/pages/changelog.astro
```diff
@@ -0,0 +1,620 @@
...
+                    class="cl-entry-toggle"
+                    type="button"
+                    aria-expanded="false"
+                    aria-controls={`${e.id}-detail`}
+                  >
+                    <span class="cl-badge" data-type={TYPE_LABEL_COLOR[e.type]}>
+                      {e.type}
+                    </span>
+                    <span class="cl-badge cl-badge-area">{e.area}</span>
+                    <span class="cl-version">{e.version}</span>
```
AbhishekDoshi26: **(suggestion)** Two things about this row:

1. Each entry has a `date`, but the page only shows the month. A small `<time datetime="2026-09-10">Sep 10</time>` next to the version would help.
2. The version has no label, so it's unclear whether it's the CLI, Flutter, or API. With a Console or API entry it gets confusing. Consider a `CLI` prefix, or making `version` optional for areas without one.

Separately, an RSS feed (`/changelog.xml` via `@astrojs/rss`) is common on changelogs and would let people subscribe.

AbhishekDoshi26:
Pushed 10ce731. It addresses the rest of my review:

- **Blocker:** replaced the example entry with six real changes from CLI 1.6.120 to 1.6.123: `--flutter-version=fvm`/`system`, the patch-load fix, Flutter 3.47.4 support, the `shorebird apps` and `shorebird channels` commands, and `shorebird patches rollback`/`rollforward`. I checked each one against the release notes and the CLI source, and each links to an existing doc page.
- **On this page:** the month headings now have ids and are passed as `headings`, so they show up there.
- **Date and version:** each entry shows its date, and the version is labeled `CLI 1.6.x`, with a release-notes link in the expanded footer.
- **RSS:** added `/changelog.xml`, linked from the page intro and from `<head>`, with a Content-Type rule in `public/_headers`.
- **PR description:** updated to point at `src/data/changelog.ts`.

Build is clean (0 errors, 0 warnings, all links valid), and cspell and `xmllint` pass.

AbhishekDoshi26:
Pushed d73ec70, which redesigns the list as a timeline:

- **Layout:** each entry's date and CLI version sit in a left column, joined by a line with a dot in the change type's color. The title and summary use the full width. At phone width it becomes one column, with the date above the badges.
- **Chips:** they show live counts that follow the search, and a chip with no matches dims. Areas with no entries (Console and API, for now) are hidden until an entry uses them. The active chip has a solid fill and `aria-pressed`.
- **Footer:** the doc link, a shorter "Release notes ↗" link, and a "Copy link" button that says "Copied" on success. It replaces the mono permalink text and the ⧉ glyph.
- **Chip row on phones:** the chips stay on one row that scrolls sideways instead of wrapping.

How I checked: `npm run build` is clean (0 errors, 0 warnings, all links valid) and cspell passes. In a local preview I checked desktop in light and dark, and phone width through a 390px frame. I measured no gaps in the timeline line, the date, dot, and badge row aligned within 1px, and no horizontal scroll at 390px. I also confirmed that the chip counts update while searching and that the live region stays quiet on load.

AbhishekDoshi26:
Pushed 6748c86, which moves each entry into its own Markdown file in `src/content/changelog/`. To add one, copy `_template.md` and fill it in. Authors write plain Markdown instead of TypeScript strings with escaped quotes.

The build now fails, with a message naming the file and line, when:
- a field is malformed: the date isn't YYYY-MM-DD, the version has a "v", or the area or type is unknown
- a CLI or Flutter entry has no `version` (it's optional for Console, API, and server-side Code Push changes, which don't ship in a CLI release)
- `docLink` points at a page that doesn't exist, or at a heading that isn't on that page
- the body has anything besides a summary paragraph, a bullet list, and one optional `sh` block

How I checked:
- `npm run build` is clean, and cspell and Prettier pass.
- `/changelog.md`, `/changelog.xml`, and the page's rendered text and attributes are byte-identical to the previous commit.
- I built 14 test entries: a copy of the template and a Console entry with no version both build, and every broken case fails with the right message and line number.

AbhishekDoshi26:
Pushed b44cb67, which makes the changelog discoverable to agents. Before this, agents could only reach it through `llms.txt`, the `.md` alternate link, and `Accept: text/markdown`:

- **`llms-full.txt`:** it now ends with the changelog. `starlight-llms-txt` only covers docs collection pages, so a build hook appends `/changelog.md` in the plugin's page format. The hook runs after the unlisted-page strip, which counts pages from the end of the file.
- **`.well-known`:** `ai-catalog.json` has a changelog entry with representative queries, `agent.json` lists it under `endpoints`, and `agent-instructions.txt` tells agents to compare an entry's CLI release with `shorebird --version` before suggesting a recently added flag, and to suggest `shorebird upgrade` if the user is behind.
- **`/changelog.md`:** it opens with the same version hint, and no longer repeats the description from its frontmatter, matching the other Markdown twins.

How I checked: the build is clean (0 errors, 0 warnings, all links valid). The build log shows the strip hook running before the append. `llms-full.txt` has 66 pages, with the changelog last. Both JSON files parse, and `changelog.xml` still validates.

src/data/changelog-schema.ts
```diff
@@ -0,0 +1,51 @@
...
+        /^\d+\.\d+\.\d+$/,
+        'version must be the Shorebird CLI release without a "v", like 1.6.123',
+      )
+      .optional(),
+    area: z.enum(AREAS),
+    type: z.enum(TYPES),
+    docLink: z
+      .strictObject({
+        label: z.string().trim().min(1),
+        href: z.string().trim().min(1),
```
AbhishekDoshi26: A relative `href` like `code-push/rollback/` passes `checkDocLink`, which resolves it against the site root. On the page, though, the browser resolves it against `/changelog/`, so it links to `/changelog/code-push/rollback/`, a 404. I'll require `href` to start with `/` or `https://`.

src/data/changelog.ts
```diff
@@ -0,0 +1,215 @@
...
+    } else {
+      fail(n, `Unexpected text: "${line.trim().slice(0, 40)}".`);
+    }
+  }
+
+  if (summary.length === 0)
+    fail(undefined, 'The summary paragraph is missing.');
+  if (bullets.length === 0) fail(undefined, 'At least one bullet is required.');
+  if (code?.length === 0) fail(undefined, 'The code block is empty.');
+  return {
```
AbhishekDoshi26: The page only understands backticks, but `/changelog.md` passes the text through as Markdown, so the two can drift silently:

- An odd number of backticks turns the rest of the line into code on the page.
- `[links](...)` and `**bold**` show as literal brackets and asterisks on the page, but render properly in the `.md`.
- A placeholder like `<id>` outside backticks is escaped text on the page, but an unknown HTML tag in the `.md`, which most renderers drop.

I'll check each summary and bullet for these outside code spans, and fail with a message that says what to do instead (use `docLink` for links, and wrap placeholders in backticks).

src/data/changelog.ts
```diff
@@ -0,0 +1,215 @@
...
+      `${file}: docLink.href "${href}" links to #${fragment}, but that page ` +
+        `has no such heading. Its headings are: ` +
+        headings.map((h) => `#${h.slug}`).join(', '),
+    );
+  }
+}
+
+/** A copy of `entries`, newest first. Same-day entries keep their order. */
+export function newestFirst(entries: ChangelogEntry[]): ChangelogEntry[] {
+  return [...entries].sort((a, b) => b.date.localeCompare(a.date));
```
AbhishekDoshi26: "Same-day entries keep their order" relies on the order `getCollection` returns files in, which isn't documented as stable. Two entries share Sep 16 and two share Sep 14, so the page and RSS order could change between builds. I'll break ties by file name and say so in the template.

src/pages/changelog.astro
```diff
@@ -0,0 +1,896 @@
...
+      });
+    }
+
+    // Open and scroll to the entry named in the URL hash, if any. Also runs
+    // on hashchange, so a permalink followed from this same page opens too.
+    function openFromHash() {
+      const targetId = decodeURIComponent(location.hash.replace(/^#/, ''));
+      if (!targetId) return;
+      const target = document.getElementById(targetId);
+      if (target?.classList.contains('cl-entry')) {
```
AbhishekDoshi26: When a search or area chip hides the target, following a permalink, or clicking a month in "On this page", does nothing. For example, search `rollback` (which only matches August), then click "September 2026". I'll clear the filters first whenever the hash names a hidden entry or month.

src/pages/changelog.xml.ts
```diff
@@ -0,0 +1,60 @@
+import { CHANGELOG_DESCRIPTION, getEntries } from '~/data/changelog';
+
+// RSS 2.0 feed of the changelog, so readers can subscribe instead of checking
+// the page. Written by hand rather than with `@astrojs/rss`, since a flat list
+// of entries needs only a few tags.
+export const prerender = true;
+
+const SITE = 'https://docs.shorebird.dev';
```
AbhishekDoshi26: `site` is already set in `astro.config.mjs`, so this can read it from the endpoint context instead of repeating the domain.

AbhishekDoshi26 (commented):
Fresh-eyes pass over the whole PR. The structure holds up. I also confirmed that `astro dev` picks up entry edits, and recovers from a broken entry once it's fixed, without a restart, so the `getEntries()` cache isn't a problem.

I found five small issues, all about things an author can write that pass the build but render wrong, plus two edge cases. I'm pushing fixes for all five right after this review.

AbhishekDoshi26:
Pushed 55bec5e, which fixes all five items from my review above:

- **Text that renders differently:** the build now fails on an unclosed backtick, a Markdown link, bold text, or a bare `<placeholder>` in a summary or bullet, and on any Markdown in a title. Each error points at the file and line and says what to do instead. The template documents the rule.
- **Relative `docLink.href`:** it must now start with `/` or `https://`.
- **Same-day order:** entries from the same day are sorted by file name. The template says so.
- **Hidden hash targets:** a permalink or month link whose target a filter is hiding now clears the filters, then opens and scrolls to it.
- **RSS links:** built from `site` in `astro.config.mjs`.

How I checked:
- The build is clean (0 errors, 0 warnings, all links valid), and cspell and Prettier pass.
- `/changelog.md`, `/changelog.xml`, and `llms-full.txt` are byte-identical to the previous commit. The page HTML is identical apart from its script.
- I built six new broken test entries (relative href, unclosed backtick, link, bold, bare placeholder, backticks in the title), and each fails with the right message and line. A copy of the template still builds.
- In a browser, I searched `rollback` and then followed the September link, and set the Flutter filter and then followed a permalink to a CLI entry. Both cleared the filters and landed on the target.

AbhishekDoshi26 (commented):
One more independent review pass. I reproduced or confirmed each finding below before posting.

The main one is about content. I seeded the entries from the GitHub release bodies, but those are auto-generated and incomplete. `RELEASE_NOTES.md` in shorebirdtech/shorebird is the full record, and it shows that 1.6.123 also shipped Flutter 3.47.5 support and an iOS `--split-debug-info` fix. I'm adding what's missing for 1.6.120 through 1.6.123: every Flutter version bump, plus the dSYM fix. Future entries should be written from `RELEASE_NOTES.md`.

I'm pushing fixes for all of these right after this review.

src/content/changelog/flutter-3-47-4.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Flutter 3.47.4 support
```
AbhishekDoshi26: This is the newest Flutter entry, but 1.6.123 shipped Flutter 3.47.5 / Dart 3.13.4 (commit "bump Flutter to ff900e7", and `RELEASE_NOTES.md`), and `src/consts.ts` already says 3.47.5. The 3.47.2 (1.6.120) and 3.47.3 (1.6.121) bumps are also missing. So a reader, or an agent following the new `agent-instructions.txt` hint, would conclude 3.47.4 is the latest. I'm adding entries for 3.47.2, 3.47.3, and 3.47.5.

src/content/changelog/failed-patch-checks-for-replacement.md
```diff
@@ -0,0 +1,20 @@
...
+
+When a patch fails to load, the device now checks for a replacement patch right
+away instead of waiting for the next launch.
+
+- Previously, reporting the failure suppressed the update check, so the device
+  ran the base release and stayed on the bad patch until it launched again.
+- Patch checks now report the patch the device is actually running, so each
+  device is attributed to the right patch.
+- These fixes are in the updater built into your app, so they apply to releases
+  built with Shorebird 1.6.122 or later.
```
AbhishekDoshi26: 1.6.122 only bumped Flutter, so this fix reaches apps through Shorebird's Flutter 3.47.4 engine, not through the CLI. Someone on CLI 1.6.122 who releases with `--flutter-version=3.44.9` won't get it (`flutter-version.mdx`: changes are not backported). I'll say "releases built with Shorebird's Flutter 3.47.4 or later".

Separately, "ran the base release" in the first bullet contradicts the linked section, which says a failed patch rolls back to the latest installed patch. I'll reword it to what the fix actually changes: the device now checks for a replacement on the same launch.

src/data/changelog.ts
```diff
@@ -0,0 +1,256 @@
...
+    } else if (/^(#|>|\d+\.\s)/.test(line)) {
+      fail(n, 'Headings, quotes, and numbered lists are not supported.');
+    } else if (line.startsWith('```')) {
+      if (code) fail(n, 'Only one code block is allowed.');
+      if (!/^```(sh|bash|shell)?\s*$/.test(line)) {
+        fail(n, 'The code block must be ```sh.');
+      }
+      code = [];
+      for (i++; i < lines.length && !lines[i].startsWith('```'); i++) {
+        if (lines[i].trim()) code.push(lines[i].trim().replace(/^\$\s*/, ''));
```
AbhishekDoshi26: This corrupts commands, and the build still passes:

- `replace(/^\$\s*/, '')` also strips the `$` from a line that starts with a variable (`$ANDROID_HOME/...` becomes `ANDROID_HOME/...`).
- `.trim()` drops the indentation of continuation lines.
- Blank lines are dropped.

I'll strip only a `$` followed by whitespace, and keep lines as written.

src/pages/changelog.astro
```diff
@@ -0,0 +1,906 @@
...
+                          {b
+                            .split('`')
+                            .map((part, i) =>
+                              i % 2 ? <code>{part}</code> : part,
+                            )}
+                        </li>
+                      ))}
+                    </ul>
+                    {e.code && (
+                      <pre class="cl-code"><code>{e.code.split('\n').map((c) => `$ ${c}`).join('\n')}</code></pre>
```
AbhishekDoshi26: Two problems with the `$ ` prompt:

- It's literal text inside `<code>`, so selecting and pasting a command gives `$ shorebird ...`, which fails in a shell.
- It's added to every line, so a `\` continuation renders as `$ --flutter-version=fvm`.

I'll render one span per line and draw the prompt with a CSS `::before`, which isn't copied, and skip it on continuation lines.

src/data/changelog.ts
```diff
@@ -0,0 +1,256 @@
...
+ * Fails the build if an internal `docLink` doesn't point at a docs page, or
+ * its `#fragment` doesn't match a heading on that page. Starlight's link
+ * validator only covers links inside docs pages, not these.
+ */
+async function checkDocLink(file: string, href: string): Promise<void> {
+  if (/^https?:\/\//.test(href)) return;
+  const url = new URL(href, 'https://docs.shorebird.dev');
+  const path = url.pathname.replace(/^\/|\/$/g, '');
+  const docs = await getCollection('docs');
+  const doc = docs.find((d) => d.id === path || d.id === `${path}/index`);
```
AbhishekDoshi26: This accepts unlisted and draft pages. `docLink.href: /zap/` builds cleanly and then advertises `/zap/` on the page, in `/changelog.md`, and in `llms-full.txt`, which is exactly what `src/unlisted.ts` exists to prevent. I'll filter them out the same way `[...slug].md.ts` does.

Two edge cases in the same function:

- An `https://docs.shorebird.dev/...` href skips validation entirely (line 165). I'll validate it as internal.
- A link to a standalone page like `/roadmap/` fails as "no docs page". I'll accept it when `src/pages/<path>.astro` exists.

src/data/changelog.ts
```diff
@@ -0,0 +1,256 @@
...
+const BODY_HELP =
+  'The body must be a summary paragraph, then a "- " bullet list, then ' +
+  'optionally one ```sh code block. See src/content/changelog/_template.md.';
+
+/**
+ * The page renders only backticks (as inline code), while `/changelog.md`
+ * passes text through as Markdown. Returns what in `text` would render
+ * differently between the two, or `undefined` if nothing would.
+ */
+function inlineProblem(text: string, allowCode = true): string | undefined {
```
AbhishekDoshi26: Some Markdown still renders differently on the page and in `.md`:

- single emphasis (`*x*`, `_x_`)
- entities (`&amp;` is double-escaped on the page)
- backslash escapes
- bare URLs, which GFM auto-links

I'll reject these too. The emphasis check requires a closing marker, so `snake_case` and `_template.md` stay allowed.

src/pages/changelog.astro
```diff
@@ -0,0 +1,906 @@
...
+    background: var(--sl-color-white);
+    border-color: var(--sl-color-white);
+    color: var(--sl-color-black);
+  }
+
+  .cl-chip.is-active .cl-chip-count {
+    background: color-mix(in srgb, var(--sl-color-black) 14%, transparent);
+  }
+
+  .cl-chip.is-empty:not(.is-active) {
```
AbhishekDoshi26: Dimming the whole chip to 45% opacity drops its text below 4.5:1 (about 3.4:1 in dark, 2.3:1 in light), and the chip is still focusable and clickable. I'll dim only the count pill, and keep the label at full contrast.

src/pages/changelog.astro
```diff
@@ -0,0 +1,906 @@
...
+          title: 'Shorebird changelog',
+          href: '/changelog.xml',
+        },
+      },
+    ],
+  }}
+  headings={groups.map((g) => ({ depth: 2, slug: g.id, text: g.label }))}
+>
+  <p>
+    What we shipped to Code Push, the CLI, and the API. Expand an entry for the
```
AbhishekDoshi26: "Code Push, the CLI, and the API" doesn't match the content: there are no API entries yet, and Flutter entries exist. The same text reaches `CHANGELOG_DESCRIPTION`, `llms.txt`, and the RSS description. I'll switch to wording that doesn't list areas.

src/pages/changelog.astro
```diff
@@ -0,0 +1,906 @@
...
+          label.textContent = 'Copy link';
+          copyBtn.classList.remove('is-copied');
+        }, 1500);
+      });
+    }
+
+    // Open and scroll to the entry named in the URL hash, if any. Also runs
+    // on hashchange, so a permalink followed from this same page opens too.
+    function openFromHash() {
+      const targetId = decodeURIComponent(location.hash.replace(/^#/, ''));
```
AbhishekDoshi26: An unguarded `decodeURIComponent` throws on a malformed hash like `#%E0`. Because `init()` calls it before registering the `hashchange` listener and the first `applyFilters()`, the whole page script dies. I'll guard it.

Also, an entry file named like a month (`september-2026.md`) would collide with that month's heading id. I'll reject those names at build time.

AbhishekDoshi26:
Pushed f953188, which fixes every item from the review above.

**Content**
- Added Flutter 3.47.2, 3.47.3, and 3.47.5 support, and the iOS `--split-debug-info` dSYM fix. That makes 10 entries covering everything notable in 1.6.120 through 1.6.123, checked against `RELEASE_NOTES.md`.
- The 1.6.122 updater entry now says the fix ships in Shorebird's Flutter 3.47.4 engine, and no longer contradicts the rollback docs.
- The description no longer lists areas that have no entries. This also covers `llms.txt` and RSS.
- Updated the PR description to say future entries should come from `RELEASE_NOTES.md`.

**Code**
- **Commands:** lines are kept as written, and only a `$ ` prompt is stripped, so `$ANDROID_HOME/...` survives. The prompt is drawn in CSS, so copying a command leaves it out, and continuation lines get no prompt.
- **`docLink`:** unlisted and draft pages are rejected, absolute `docs.shorebird.dev` links are validated, and standalone pages like `/roadmap/` are accepted.
- **Entry text:** italics, HTML entities, backslash escapes, and bare URLs are now rejected too. `snake_case` and `_template.md` still pass.
- **File names:** a name that matches a month heading's anchor is rejected.
- **Empty chips:** they're dimmed with a dashed border instead of opacity. The labels now measure 12.85:1 in dark and 10.3:1 in light, and the prompt is 4.84:1 or better.
- **Hash links:** a malformed hash no longer stops the page script.

How I checked:
- The build is clean (0 errors, 0 warnings, all links valid), and cspell and Prettier pass. The page, `.md`, and RSS each list 10 entries, and the XML validates.
- I built test entries for every new check. Each broken case fails with the right `file:line` message, and the valid ones (`snake_case`, `/roadmap/`, and a multi-line command with `$VARIABLE`) build.
- In a browser, copying a command gives the text without `$`, and I measured the contrast values above.

PR #680, opened 2026-09-19
Style the sidebar badge like the console's soft badge

easymac:
The CI group's "deprecated" badge rendered as Starlight's default: a square mono-font chip flush against the label and hanging below its baseline, so the padding read as space inside the badge.

This ports the console's soft badge recipe from `packages/ui/src/badge.tsx`: one source color drives the text, a 35% border, and a 12% background, using the console's warning token values via `light-dark()`. The badge is now a pill in the sans font at 11px, with a gap after the label, and the group label is flex-centered so the badge lines up with the text and caret. The label is capitalized to match console badge casing.

Verified in both themes on the dev server.

cc @AbhishekDoshi26


src/styles/custom.css
```diff
@@ -82,3 +82,25 @@ h6 {
...
+.sidebar-content .sl-badge {
+  --badge-color: light-dark(
+    oklch(0.7517 0.1106 52.07),
+    oklch(0.8208 0.1259 74.91)
+  );
+  margin-inline-start: 0.5rem;
+  border-radius: 9999px;
+  padding: 0.125rem 0.5rem;
+  font: 500 0.6875rem/1.4 var(--sl-font);
+  color: var(--badge-color);
```
AbhishekDoshi26: Suggested fix (tested in both themes: 5.11:1 in light, dark unchanged):

```suggestion
  /* Light text uses the console's `--on-warning`: `--warning-500` on its own
     tint over white is 2.1:1, under WCAG AA's 4.5:1 for small text. */
  color: light-dark(oklch(0.5264 0.1135 48.92), var(--badge-color));
```

AbhishekDoshi26 (changes requested):
Nice cleanup. The pill reads much better than the square chip, the label, badge and caret now share a center line (measured to within 0.01px), and I checked that `light-dark()` follows the site's theme toggle, which sets `color-scheme` along with `data-theme`. The port is faithful to `softSurfaceClasses` and the console's `--warning-500` values.

One blocker: **the light-mode badge text fails contrast.** `--warning-500` (`#e69a6b`) on its own 12% tint over the white sidebar is **2.08:1**. WCAG AA needs 4.5:1 for 11px text, and production is currently at 6.36:1, so this is a regression. Dark mode is fine at 8.61:1. All three numbers were measured with computed styles in the browser.

The console's tokens already include the fix: `--on-warning` (`oklch(0.5264 0.1135 48.92)`) is its text-safe warning color for light surfaces. Using it for the text in light mode only gives **5.11:1**, while the border and fill keep the soft recipe and dark mode stays unchanged. I tested the suggestion inline below in both themes. (The console's own "Needs attention" soft badge in `plan-card.tsx` has the same 2:1 problem, which may be worth passing along.)

Non-blocking: `.sidebar-content .sl-badge` applies the warning color to every sidebar badge. Only CI has one today, but scoping it to `.sl-badge.caution` would stop a future `note` or `success` badge from rendering orange.

CI is green.

AbhishekDoshi26 (approved):

